PR #1, opened 2016-05-24
Add consolidate_branches script

guewen:
It allows to create branches with a selection of pull requests. This
selection comes from a configuration file so we can regenerate the
consolidated branch each time a pull request is added, removed or added.


guewen:
won't work from inside the docker container as it needs `git` to be installed and the correct credentials...

guewen:
Replaced by https://github.com/acsone/git-aggregator

PR #7, opened 2016-07-13
Fix error pkg_resources.DistributionNotFound: odoo==9.0c

guewen:
The build runs `pip install -e` on the odoo src, which creates an
odoo.egg-info directory *inside /opt/odoo/src*. So when we run a
container with a volume shared with the host, we don't have this
.egg-info (at least the first time).

Correction: When it happens, we reinstall the odoo package. We don't want to run it anytime
because it would slow the start of the containers


mdietrichc2c:
:+1:

PR #11, opened 2016-09-06
[ADD] xlsxwriter pypi module for account_financial_report_qweb

vrenaville:


yvaucher:
Do we want this dependency everywhere ? We now have a bunch of project without accounting thus won't need xlsxwriter

guewen:
I agree with @yvaucher, we should add it in the projects only if necessary.

guewen:
In Odoo 10, it becomes a required dependency, so I added it for Odoo 9.0 as well in #12.

PR #17, opened 2016-12-13
Add DB_HOST environment variable

jstoja:
Hello,

It might be convenient to be able to set the DB_HOST variable instead of specifying the extra_host when launching.
It's little config and retro-compatible with the default value.

### DIFF
--- a/10.0/etc/odoo.cfg.tmpl
+++ b/10.0/etc/odoo.cfg.tmpl
@@ -2,7 +2,7 @@
 addons_path = {{ .Env.ADDONS_PATH }}
 data_dir = /data/odoo
 auto_reload = False
-db_host = db
+db_host = {{ default .Env.DB_HOST "db" }}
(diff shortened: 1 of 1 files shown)


guewen:
Hello, thanks for the pull request!

Don't you use `docker-compose.yml`? The extra host has to be set only once if you do.

Actually, it's not retro-compatible as we use it in various scripts and we might have used the `db` hostname in "sub-images". So I'm not too fond of changing it unless we have a very good reason.

Plus, I'm not sure that having two different ways of setting the db hostname is for the better.

guewen:
Yes you are right for the sub-images.

Still, between an environment variable and the extra hosts, depending on the tools you use, the latter allows to be more flexible, for instance in Rancher you can do the following to use a container of another composition:

```
  external_links:
    - postgres-cluster/lb:db
```

guewen:
Hey, I completed this feature on #25. You might want to check.
Thanks

PR #19, opened 2017-01-18
Bump anthem version

yvaucher:


guewen:
Could you update the history file?

yvaucher:
History.rst updated

PR #20, opened 2017-01-19
[add] set "report.url" param too

simahawk:


guewen:
We have a use case where we want a different url for reports than for the base url. I suggest to add `ODOO_REPORT_URL` that fallbacks to `$DOMAIN`.

vrenaville:
+1 for the report.url

PR #21, opened 2017-02-01
install required dependencies for xls/xlsx/ods imports

gurneyalex:


guewen:
Shouldn't them be added in 9.0 as well?

gurneyalex:
@guewen you're right, I've updated this.

guewen:
Thanks

PR #23, opened 2017-04-04
[add] support extra dev requirements

simahawk:


9.0/Dockerfile
```diff
@@ -2,11 +2,10 @@ FROM debian:jessie
 MAINTAINER Camptocamp
 
 # create the working directory and a place to set the logs (if wanted)
-RUN mkdir -p /opt/odoo /var/log/odoo
+RUN mkdir -p /opt/odoo/requirements /var/log/odoo
 
 WORKDIR "/opt/odoo"
-
-COPY ./base_requirements.txt ./
+COPY ./requirements/*requirements.txt ./requirements/
```
guewen: `COPY ./requirements/ ./requirements/`

9.0/Dockerfile
```diff
@@ -2,11 +2,10 @@ FROM debian:jessie
 MAINTAINER Camptocamp
 
 # create the working directory and a place to set the logs (if wanted)
-RUN mkdir -p /opt/odoo /var/log/odoo
+RUN mkdir -p /opt/odoo/requirements /var/log/odoo
```
guewen: change not needed if we copy the directory as I propose below

9.0/Dockerfile
```diff
@@ -104,4 +102,4 @@ ONBUILD COPY ./setup.py /opt/odoo/
 ONBUILD COPY ./VERSION /opt/odoo/
 ONBUILD COPY ./migration.yml /opt/odoo/
 # need to be called at the end, because it installs . and src
-ONBUILD RUN cd /opt/odoo && pip install -r src_requirements.txt
+ONBUILD RUN cd /opt/odoo && find . -maxdepth 1 -name "*requirements.txt" -exec pip install -r {} \;
```
guewen: It won't install `src_requirements.txt` isn't it? Shouldn't it be:

```
ONBUILD RUN cd /opt/odoo && pip install -r requirements/src_requirements.txt && find . -maxdepth 1 -name "*requirements.txt" -exec pip install -r {} \;
```

simahawk: oops, right :)

9.0/Dockerfile
simahawk: +1

simahawk:
as we agreed: I pushed this to [dev_requirements](https://github.com/camptocamp/docker-odoo-project/tree/dev_requirements) branch to produce a testable image.

simahawk:
REMINDER: if this works update the example project too and its readme

guewen:
I don't remember what was the status on this, does it work?

simahawk:
my test failed. Sounds like extra requirements are not there when they should ONBUILD.
Unless it works for you I'd say we should drop this and keep it only in the project setup which works for sure.

guewen:
> Unless it works for you I'd say we should drop this and keep it only in the project setup which works for sure.

:+1: Thanks for trying :)

PR #33, opened 2017-07-06
Use debian stretch as base OS

guewen:


yvaucher:
Restarting travis job as we had bunch of :
503  Service Temporarily Unavailable

guewen:
we have issues with the wkhtmltopdf which is packaged only for jessie by odoo

yvaucher:
https://github.com/odoo/odoo/issues/15308#issuecomment-276934340

According to sle-odoo only Wheezy and Jessie are supported for Odoo 10.0

Thus I guess we better wait for version 11.0 before we change this.

guewen:
interesting, thanks for the info

guewen:
Done in 2.5.0 with 11.0 support.

PR #37, opened 2017-09-08
Enable publishing of maintenance web page

guewen:
By upgrading anthem to 0.10.0 and marabunta to 0.7.0.


guewen:
A new anthem release with https://github.com/camptocamp/anthem/pull/25 will be needed.

PR #43, opened 2017-11-28
Add future on images with python 2.7 to be able to have cross python …

yvaucher:
…tools

This prepare the arrival of odoo v11


jcoux:
@yvaucher Locally, it works for me.
But we need to activate `python-dev` and `build-essential` for the time where we install requirements files.

yvaucher:
@jcoux python-dev and build-essential are uninstalled after we do the pip install:

https://github.com/camptocamp/docker-odoo-project/blob/master/10.0/Dockerfile#L53

PR #50, opened 2018-02-21
10.0 update py deps

yvaucher:
https://github.com/odoo/odoo/blob/aad677c338ccd50c14e7a10314e3d18e76fe3f88/requirements.txt#L1


yvaucher:
```
-chardet==2.3.0
-colorama==0.3.2
-future==0.16.0
-gdata==2.0.18
-html2text==2016.9.19
-html5lib==0.999
-odfpy==1.3.4
-odoo-autodiscover>=2.0.0b1
-pyinotify==0.9.4
-python-stdnum==1.0
-roman==2.0.0
-simplejson==3.6.5
-urllib3==1.9.1
-unittest2==0.5.1

+jcconv==0.2.3
+pydot==1.2.3
+ebaysdk==2.1.4
```

yvaucher:
I'll restore future and simplejson, for the others I didn't find any relvent module that used them directly as dependency.

guewen:
> -odoo-autodiscover>=2.0.0b1

is required for pip installing addons



> -pyinotify==0.9.4

required for hot reloading of assets, well it doesn't work with docker volumes...



> -urllib3==1.9.1

dependency of requests

> -html2text==2016.9.19

It's in odoo/tools/misc.py

> odfpy

used by base_import



Contrariwise, I'm pretty sure we can remove simplejson which is now a std (import json). 

We need to double-check this list.

yvaucher:
I'll restore those dependency adding a better comment on them to identify their use, like the odoo-autodiscover which was added lately for pip install. It's not a direct dependency of odoo

yvaucher:
html2text is used to format an alternative body in emails for reader which couldn't read HTML.

Strange though that there is a lib inside odoo in addons/mail/models/html2text.py

guewen:
To replace by #65

PR #55, opened 2018-03-29
Disable pip's cache dir to reduce size of images

### LINKED ISSUE #51: Don't store pip cache dir
The `--no-cache-dir` option should be added to the `pip install` commands in Dockerfiles to reduce the size of the images.

See https://pip.pypa.io/en/stable/reference/pip_install/#caching

guewen:
Fixes #51

### DIFF
--- a/install/setup-pip.sh
+++ b/install/setup-pip.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+set -eo pipefail
+
+cat << EOF > /etc/pip.conf
+[global]
+# counter-intuitively, false means that we enable 'no-cache-dir'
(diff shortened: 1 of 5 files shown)


install/setup-pip.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -eo pipefail
+
+cat << EOF > /etc/pip.conf
+[global]
+# counter-intuitively, false means that we enable 'no-cache-dir'
```
simahawk: WTF???? :smile:

p-tombez: Who designed this thing :interrobang:

PR #57, opened 2018-03-29
Fix permission issue when installing runtests db

guewen:
By consistency, in the commit 0c58332 I changed the path to the odoo-bin
script in both places it was used.
However, as long as 'coverage run' accepts a file which doesn't to be
executable, the first call is executed. We could call python2 or python3
on ${ODOO_BIN_PATH}, but the "odoo" script is there for that. So let's
keep "odoo" for the install and use the "odoo-bin" only for 'coverage
run' where it's required.

For the record, the 'odoo-bin' script is eXecutable in the git
repository, but in some cases we optimize the build of the image by
downloading a tar, and it seems the x flag is absent there.

### DIFF
--- a/bin/runtests
+++ b/bin/runtests
@@ -32,7 +32,7 @@ fi
 DB_NAME_TEST=${DB_NAME}_test
 
 PGPASSWORD=$DB_PASSWORD createdb -h $DB_HOST -U $DB_USER -O $DB_USER ${DB_NAME_TEST}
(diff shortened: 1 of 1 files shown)


yvaucher:
not executable in tar archive? damn, that's a serious issue if permissions are not preserved. Shouldn't we consider to always do a chmod +x on that file?

guewen:
@yvaucher I don't think it's necessary as we should always use the `odoo` executable which is available in the $PATH. Here I only wanted too much consistency but really there was no reason to call `odoo-bin` directly.

PR #58, opened 2018-04-03
Add 'unaccent' support

damdam-s:


simahawk:
I don't see the pg contribs been installed. Aren't they needed for this feature to work?

damdam-s:
@simahawk I don't know if this is managed in the docker postgresql image. If the extension is available, this should work.
I'm trying to test it locally

guewen:
It's good to have in any case here, the postgres contrib is part of another container or setup.

simahawk:
hmm, then would be nice to have a comment saying "hey, dude, you need to install this and this or use another image to use this",  no?

guewen:
why not, the point is we don't have a "standard" postgres image, so the doc mostly refers to odoo's documentation, but it wouldn't hurt to add a link to the odoo documentation explaining the requirements

PR #61, opened 2018-06-18
Refactor image, split in 2 (with and without onbuild) with 2 version (light python package, full python package)

sebastienbeau:
Hi as discuss with @guewen, the aim of this PR is to:
- [x] have a base image without onbuild feature. As when we mount directory in the image we may not want to copy all code
- [x] refactor code in order to share more logic between image, and add a global variable with "dev/build package"
- [x] add a full version image with all python dependency needed for OCA and futur OCA module
- [x] ~~pre-install odoo with dev mode and remove the source (in order to generate in the image the python bin and path for odoo)~~
- [x] to discuss give the possibility to mount the whole directory odoo and not only each path manually
- [x] update example base on onbuild image and base image


sebastienbeau:
build failed due to pip package issue : https://github.com/pypa/pip/issues/5221

Makefile
```diff
@@ -19,27 +19,32 @@ build:
 	cp -r start-entrypoint.d/ $(TMP)
 	cp -r before-migrate-entrypoint.d/ $(TMP)
 	docker build --no-cache -t $(IMAGE_LATEST) $(TMP)
+	docker build --no-cache -f $(TMP)/Dockerfile-onbuild -t $(IMAGE_LATEST)-onbuild $(TMP)
```
guewen: This change is a bit risky for us as our projects Dockerfiles will start using the images without onbuilds if we don't take care, should we prefer to add the suffix to the one without onbuilds like `-core`? On the other hand, the new names will be used only for the next release (or latest that we shouldn't use) and at this point if we upgrade to the new release without adding the `-onbuild`, the tests won't pass so we'll know... and I'd prefer to have `-onbuild` rather than `-core`. So... any opinion on that @yvaucher @simahawk?

10.0/Dockerfile-onbuild
```diff
@@ -0,0 +1,15 @@
...
+ONBUILD COPY ./src /opt/odoo/src
+ONBUILD COPY ./external-src /opt/odoo/external-src
+ONBUILD COPY ./local-src /opt/odoo/local-src
+ONBUILD COPY ./data /opt/odoo/data
+ONBUILD COPY ./songs /opt/odoo/songs
+ONBUILD COPY ./setup.py /opt/odoo/
+ONBUILD COPY ./VERSION /opt/odoo/
+ONBUILD COPY ./migration.yml /opt/odoo/
+# need to be called at the end, because it installs . and src
+ONBUILD RUN cd /opt/odoo && pip install -r src_requirements.txt
```
guewen: Curious about this one though, does it mean that the libs will be installed only when you start the container?

Makefile
sebastienbeau: On my side using -onbuild seem more clean, but for compatibility issue I do not see any problem of using -core, but it's true that when you will update your test will fail so maybe there is no risk to use '-onbuild'

10.0/Dockerfile-onbuild
sebastienbeau: For now yes, but as most of lib are already here, so it's very fast as it just run a pip install for installing odoo in dev mode. So it work but not very clean solution

After inspecting the docker container the pip install add this 3 files in the image:
/usr/local/lib/python2.7/dist-packages/easy-install.pth
/usr/local/lib/python2.7/dist-packages/odoo.egg-link
/usr/local/bin/odoo

and for sure also add the odoo.egg-info in src folder (but this is mounted)

After some test I see to possibility to "pre-installing odoo".
- adding this 3 files
- downloading odoo (in a RUN cmd) installing it in dev mode and removing it.

I can do it here or it a specific akretion image based on "-core" image. What do you think?

Makefile
yvaucher: No strong opinion on this, we anyway have a new layer so a big change, the base layer being usable on it's own with volumes or with additionaly actions in a Dockerfile. I'm not chocked if this is a breaking change. This needs to be documented though.

The main risk here is someone using `latest` tags.

It is true that a developer who is unaware of this change could run fine a container locally due to his volume configuration and won't expect the final image to fail as it will be suddently missing all `ONBUILD`.

Nevertheless, it is better to change that early, and the `-onbuild` makes more sense.

On dockerhub I see only 5 stars (1 from me) and 10k download with all the CI we have it covers most of the downloads I woud say.
Thus the impact should be minimal.

If you think there is a risk to affect many people, we could still detect a bad use of the new splitted images.
An idea would be to check the presence of a specific file or env var we insert in `-onbuild` image and if not present we show a WARNING message with an entrypoint script explaining the 2 images and their purpose. The warning could be silenced by adding the same filename or env var in your custom Dockerfile if you don't want to use the `-onbuild` image. Thus, even with volumes you would be warned it has been splitted and you have the choice to ignore or to use the `-onbuild` image to restore previous behavior. Before you build and push a new image, I expect it to be at least once tested locally.

install/package_p2.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
sebastienbeau: Maybe I can named it package_odoo_9.0_10.0.sh as it the same?

sebastienbeau:
Ok it seem that changing the lastest with the version without onbuild will not break a lot of thing as most of user are C2C team. Also as the good practice is to never use the lastest. I think we can do the move.

install/package_p2.sh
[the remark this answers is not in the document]
yvaucher: Naming it for each version would need to imply to change package_odoo_11.0 to package_odoo_11.0_12.0 and so on. That's why I propose the first version for which it applies.

Makefile
guewen: Agree to have the `-onbuild`. Nobody should use latest ;) Well if we have to add warning, we could simply detect if we have some content in the `src` directory, if you have not it means you have either no `onbuild` or no volumes, but not sure it's required.

install/preinstall_odoo.sh
```diff
@@ -0,0 +1,11 @@
...
+set -eo pipefail
+
+ODOO_URL=https://github.com/odoo/odoo/archive/$ODOO_VERSION.tar.gz
+
+curl -o /tmp/odoo.tar.gz -SL $ODOO_URL
+tar xfz /tmp/odoo.tar.gz -C /opt/odoo/
+mv /opt/odoo/odoo-$ODOO_VERSION/ /opt/odoo/src
+pip install -e /opt/odoo/src
+rm -f /tmp/odoo.tar.gz
+rm -rf /opt/odoo/src
```
simahawk: overall looks good to me, just a question about this: am I wrong or this is going to run every time, dev mode included?
How it deals w/ forked/aggregated odoo/src via submodule?
Also, wouldn't be better to make `ODOO_URL` a global var so that we can override it IF needed?

sebastienbeau: In fact it do not really install odoo. It only install it (with pip dev mode) and remove the source. There is not odoo code in the image, you only have the python bin and the link to the source code.
So basically when you are in dev and you start your docker the first time
- the entrypoint will generate the odoo.egg in your odoo directory
- the link in lib and the odoo bin already exist (nothing done here)

Then if you restart your docker, as the odoo.egg already exist and as bin and lib link already exist, it work without using onbuild !

Note it will be also possible to not preinstall odoo but this mean installing it in the Dockerfile of your project. But as odoo code source is huge, building the context take time and I do not want to wait !

sebastienbeau:
@guewen @simahawk @yvaucher I have finish my last change, now I am going to battle test it on my side. Tell me if you are agree with my last change

10.0/extra_requirements.txt
```diff
@@ -1,8 +1,12 @@
 # Extra python dependencies
 algoliasearch==1.17.0
+Adyen==1.2.0
+cachetools==2.1.0
+cerberus==1.2
 boto3==1.7.40
 factur-x==0.3
 invoice2data==0.2.74
+magento==3.0
```
simahawk: hmmm do we need these packages in base setup? :)
was meant to stay in example folder maybe?

example/docker-compose.override.yml
```diff
@@ -13,12 +13,8 @@ services:
       - 8072
     volumes:
       - "data-odoo-pytest-cache:/opt/odoo/.cache"
-      - "./odoo/src:/opt/odoo/src"
-      - "./odoo/local-src:/opt/odoo/local-src"
-      - "./odoo/external-src:/opt/odoo/external-src"
-      - "./odoo/songs:/opt/odoo/songs"
-      - "./odoo/migration.yml:/opt/odoo/migration.yml"
-      - "./odoo/data:/opt/odoo/data"
+      # to spead build of image customise your .dockerfileignore (this is an help their)
```
simahawk: *speed

example/odoo/Dockerfile-onbuild
```diff
@@ -0,0 +1,15 @@
+FROM camptocamp/odoo-project:10.0-latest-onbuild
```
simahawk: worth to mention when this option shall be used?

simahawk: hmm maybe better in common/Dockerfile-onbuild

10.0/extra_requirements.txt
sebastienbeau: I have two requirements, the base_requirements and the extra_requirement (this is a new file introduced y this PR). The aim of the extra_requirement is to build a full version of the image with most used python lib (basically OCA python dependency + some extra)
You will have this lib only in the "FULL" image (ex: 10.0-full-latest, 10.0-latest-onbuild). If you do not want to have this lib preinstalled, just use the normal image ex : 10.0-latest or 10.0-latest-onbuild

FULL image can be used in two case:
- fast demo (no need to search all dependency)
- centralized python lib dependency (you want to have the same version of package for the majority of your project, so you always use the FULL version and add missing lib there. For the project where you want to manage the dependency yourself, just use normal image).

example/docker-compose.override.yml
sebastienbeau: fixed

example/odoo/Dockerfile-onbuild
sebastienbeau: Adding a section in the readme that explain the difference between the base version and the onbuild version? I can also explain the difference between the full version and normal version?

simahawk: the more you explain the better! thanks :heart:

README.md
```diff
@@ -20,6 +20,45 @@ $ make VERSION=11.0  # generate image camptocamp/odoo-project:11.0-latest
...
 ```
+## Image version
+
+There is 4 versions of image:
+- odoo-project:${odoo_version}-${tag_version}
+- odoo-project:${odoo_version}-${tag_version}-onbuild
+- odoo-project:${odoo_version}-${tag_version}-full
+- odoo-project:${odoo_version}-${tag_version}-full-onbuild
+
+Note: in production, docker hardly recommanded to use never use latest tag but instead use a specific version in order to be able to rebuild identicall images.
```
simahawk: fix typos and repetitions: "Note: in production, docker docs strongly recommend to never use "latest" tags. Instead use a specific version in order to be able to rebuild identical images."

README.md
```diff
@@ -20,6 +20,45 @@ $ make VERSION=11.0  # generate image camptocamp/odoo-project:11.0-latest
...
+
+The dockerfile of onbuild image is here:  [common/Dockerfile-onbuild](common/Dockerfile-onbuild)
+
+
+Two example of Dockerfile have been proposed in the project example here:
+
+- using normal version: [example/odoo/Dockerfile](example/odoo/Dockerfile)
+- using onbuild version: [example/odoo/Dockerfile-onbuild](example/odoo/Dockerfile-onbuild)
+
+Note: Onbuild version is shorted but
```
simahawk: *shorter / faster?

sebastienbeau: the dockerfile of the Onbuild version is shorter but ;)

sebastienbeau:
@simahawk I have done the change

README.md
```diff
@@ -20,6 +20,45 @@ $ make VERSION=11.0  # generate image camptocamp/odoo-project:11.0-latest
 $ make VERSION=10.0  # generate image camptocamp/odoo-project:10.0-latest
 $ make VERSION=9.0  # generate image camptocamp/odoo-project:9.0-latest
 ```
+## Image version
+
+There is 4 versions of image:
```
rvalyi: s/is 4 versions of image/are 4 flavors of the image/

sebastienbeau: done

sebastienbeau:
@guewen I have updated the change log, for the conflix, do you want me to solve it, or you are going to do it?

guewen:
> I have updated the change log, for the conflix, do you want me to solve it, or you are going to do it?

I can take care of it. Thanks for the changelog update!

sebastienbeau:
Thanks Guewen

guewen:
@sebastienbeau I released 2.7.0, the last release before 3.0.0, I plan to resolve and merge on Monday (and hopefully make the release 3.0.0 before I go on vacations on Wednesday)

guewen:
Rebased and merged at https://github.com/camptocamp/docker-odoo-project/commit/a259748f0fcc3fe79e72a718afee5de7313f9750 :fireworks:

PR #62, opened 2018-06-19
Wkhtmltopdf 0.12.5.1

damdam-s:
superseeds #59

### DIFF
--- a/10.0/Dockerfile
+++ b/10.0/Dockerfile
@@ -49,8 +49,8 @@ RUN set -x; \
         && apt-get install -y --no-install-recommends \
             postgresql-client \
             libpq-dev \
-        && curl -o wkhtmltox.deb -SL https://builds.wkhtmltopdf.org/0.12.5-rc/wkhtmltox_0.12.5-0.20180529.138.rc~37e9f91~jessie_amd64.deb \
-        && echo '390ae8b32c527eac1ae52b3a6b3157367adbb14e wkhtmltox.deb' | sha1sum -c - \
+        && curl -o wkhtmltox.deb -SL https://downloads.wkhtmltopdf.org/0.12/0.12.5/wkhtmltox_0.12.5-1.jessie_amd64.deb \
+        && echo '4d104ff338dc2d2083457b3b1e9baab8ddf14202 wkhtmltox.deb' | sha1sum -c - \
         && dpkg --force-depends -i wkhtmltox.deb \
         && apt-get -y install -f --no-install-recommends \
         && pip install -U pip setuptools && pip install -r base_requirements.txt \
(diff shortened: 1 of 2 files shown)


yvaucher:
For the first travis job error this might help: #64

guewen:
I read at several places than upgrading wkhtmltopdf changed the layout of many reports, we may want to be extra-cautious then

leemannd:
@damdam-s Is it still needed?

damdam-s:
@leemannd I have this in prod for a customer and seems to work pretty good. I don't know if we want this version of wkhtmltopdf in the base image.
@guewen opinion?

guewen:
> @guewen opinion?

Do you know if the printed invoices did changed? I read in many places that the scale changed a bit in this version and some reports would then be offset

damdam-s:
i'm not informed by the customer of any bug or layout issue on other reports

guewen:
Oh, it seems the issues I mentioned are being fixed here https://github.com/odoo/odoo/pull/28864

beledouxdenis:
@guewen 
I am glad my pull request can help you out :). It was becoming a technical debt for us as well to be stuck to wkhtmltopdf 0.12.1.

@damdam-s 
Up to now, using wkhtmltopdf 0.12.5 instead of 0.12.1 did render the reports with a different zoom level. Meaning, if you print the same report in Odoo (10.0/11.0/12.0) with 0.12.1 and 0.12.5 (even 0.12.2) you will see a difference in the zoom level.

My pull requests, which is being merged by our merge bot at the time I write this comment, aims to bring the compatibility so the zoom level remains exactly the same. I spent hours to find the trick.

That being said, for your customer, as you used 0.12.5 before we claimed it as officially supported, if you update the source code so it includes this compatibility revision, the reports of your customer will have a different zoom level, obviously, as it will now apply a zoom in (96/90 = 1.066667). If you want to keep your former zoom level, you can simply set "96" in your reports paper format: This will apply a zoom of 1.0, in other words no zoom.

guewen:
@beledouxdenis Thanks a lot for the heads-up here, and congrats for finding the trick :hugs:

yvaucher:
@damdam-s can you rebase / add an entry in changelog and apply the same change to version 12.0 ?

guewen:
As we need a patch / recent revision on Odoo, should we always use the new version on 9.0/10.0/11.0?

bealdav:
Hi,

As seen in #115 we have a real bug with 0.12.4 for v12 in one report.
No bug with runbot odoo v12, as i suppose it use 0.12.5.1 as supported version like 0.12.1

I plan to make a script wkhtml_12_5.sh called only in v12 Dockerfile for now.

What do you think ?

guewen:
> I plan to make a script wkhtml_12_5.sh called only in v12 Dockerfile for now.

Agree

bealdav:
Done in #115 
Can be closed here

PR #65, opened 2018-07-26
Upgrade python libs

guewen:


yvaucher:
This might create some more conflict in https://github.com/camptocamp/docker-odoo-project/pull/61
though.

guewen:
My plan is to create a release 2.6.2 with these upgrades, because several libs have a CVE.
#61 will be release 3.0.0. BTW we should not have any conflict because #61 doesn't change the base libs.

PR #66, opened 2018-07-30
Update example

guewen:
* Use version 11.0
* docker-compose.yml is now an "all-included" composition for dev, the
split with '.override' is too complicated
* consequence of the change above, the Makefile uses another
'test-compose.yml', which does not include the host volume './odoo' to
assert that the content is copied in the image and not only read from
the host volume
* upgrade nginx image, 1.2.x are broken


guewen:
@sebastienbeau @yvaucher Updated, I extracted the build and tests scripts which are now much easier to read (and change) than in the Makefile

PR #68, opened 2018-07-31
Reorganize directory structure

### LINKED ISSUE #67: move /opt/odoo to /odoo
We may want to move /opt/odoo to /odoo, some reasons

- when we work from the outside of a container in a project, we have the directory "odoo" at the root directory, so the same structure inside the container is less confusing, and means we don't need to add/remove /opt in scripts
- we anyway already have /data/odoo which isn't an usual unix directory and it seems it's not so important in the docker world to use our own directory in / (?)

Things to care about:

- adapt the paths in the container
- the directories `/opt/bin` and `/opt/etc` should be moved elsewhere to not be shadowed when we use a host volume on the full `/odoo` directory

guewen:
* /opt/odoo/etc/odoo.cfg.tmpl → /templates/odoo.cfg.tmpl
* /opt/odoo/etc/odoo.cfg → /etc/odoo.cfg
* /opt/odoo → /odoo
* /opt/odoo/bin → /odoo-bin
* /opt/odoo/bin_compat → /odoo-bin-compat (for 9.0)
* /opt/odoo/before-migrate-entrypoint.d → /before-migrate-entrypoint.d
* /opt/odoo/start-entrypoint.d → /start-entrypoint.d

This way we have the same path ./odoo outside and inside of the
container.

bin, etc and entrypoints are not in /odoo, so we can share the whole
"odoo" directory with a host volume without shadowing them.

Closes #67


guewen:
I have yet to test it on a project :)

guewen:
> I have yet to test it on a project :)

Forgot to say I could test it on a project

PR #70, opened 2018-08-08
WIP: OpenERP 6.1 support: back to the future!

rvalyi:
WORK IN PROGRESS

you can typically test it with:

```
docker run --name some-postgres -e POSTGRES_PASSWORD=odoo -e POSTGRES_USER=odoo -d postgres:9.6.1
docker run -v /root/aflex-tests/odoo:/opt/odoo -e MIGRATE=False -e ADDONS_PATH=/opt/odoo/src/addons --link some-postgres:db -ti camptocamp/odoo-project:6.1-latest openerp-server
```

### DIFF
--- a/6.1/Dockerfile
+++ b/6.1/Dockerfile
@@ -1,0 +1,81 @@
+FROM debian:jessie
+MAINTAINER Akretion
+
+# create the working directory and a place to set the logs (if wanted)
+RUN mkdir -p /opt/odoo /var/log/odoo
+
+WORKDIR "/opt/odoo"
+
+#RUN apt-get update && apt-get install -y --no-install-recommends apt-transport-https || echo "failed"
+#RUN apt-get install -y --no-install-recommends apt-transport-https
+#COPY ./install/apt.conf /etc/apt/apt.conf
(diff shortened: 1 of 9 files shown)


rvalyi:
note: with this PR and the new `bin/docker-entrypoint_6.1.sh` entry point we typically only start the main process but not Gunicorn. Both processes would typically be started with supervisor. However, in our project we used Ruby Bluepill because it seems it was easier for us to properly supervise the memory usage of the broken libreoffice we used for Aeroo reports (I'll eventually point to the proper Akretion branch later). But all I mean is that supervisor support was out of scope of this simple PR.

yvaucher:
@rvalyi do you still have interest in this?

gurneyalex:
I'm :-1: on maintaining python2 support on this project.

rvalyi:
Hello, no more interest for this PR on my side.
About more recent Python 2.7 containers: we have some interest to migrate a few V8 customers, so for v9 and v10 (possibly just this year). But we will do it our way, I understand you deprecate theses versions.

gurneyalex:
FTR we are in the progress of a major change to stop running as root, and we will likely drop support for python2 in the process.

PR #73, opened 2018-09-06
Upgrade marabunta to 0.9.0

guewen:


simahawk:
a release note to not miss: you must update odoo-template too otherwise you won't be able to release if you go w/ 5digits

guewen:
> a release note to not miss: you must update odoo-template too otherwise you won't be able to release if you go w/ 5digits

Anyway we have to provide a small internal migration guide for switching to 3.0.0, there are a few little changes to do in the existing projects.

PR #79, opened 2018-10-11
Unfreeze pluggy version

yvaucher:
pytest 2.9 requires pluggy>=0.7

By fixing pluggy we broke pytest.

This fixes failure on python TypeError: __call__() got an unexpected keyword argument 'warn_on_impl'

### DIFF
--- a/10.0/base_requirements.txt
+++ b/10.0/base_requirements.txt
@@ -67,6 +67,7 @@ anthem==0.12.0
 # test / lint
 flake8
 pytest
+pluggy
 pytest-odoo
 coverage
 pytest-cov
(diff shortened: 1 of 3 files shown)


10.0/base_requirements.txt
```diff
@@ -67,6 +67,7 @@ anthem==0.12.0
 # test / lint
 flake8
 pytest
+pluggy
```
simahawk: for our future selves, I'd add a comment on top of it saying: keep this unpinned to satisfy `pytest`

yvaucher: I'll add a comment on the # test / lint line why it is unpinned, everything in this section is under the same rule.

yvaucher: @simahawk fixed

PR #80, opened 2018-10-11
Fix BEFORE_MIGRATE_ENTRYPOINT_DIR & START_ENTRYPOINT_DIR to remove /odoo

Tonow-c2c:
To fix the problem with this parts on Dockerfile from odoo-template Line 33 and 34


yvaucher:
@Tonow-c2c your links are on a private repository, this one is public.

PR #82, opened 2018-10-19
Add release documentation to Readme

Tonow-c2c:


leemannd:
Maybe we want to change the readme and add section into the beginning of the file.
Because it is pretty long.

PR #87, opened 2018-11-07
no_database_list is not an option

sbidoul:
list_db is, but AFAICT is only a command line option, it's not supported in the config file?


guewen:
Absolutely, the `dest` or `--no-database-list` is `list_db`. Thanks!

guewen:
Thanks, merged at 8f7a8608063bb5a879b8c5172ddd71fd9679094f

PR #91, opened 2018-12-18
Add reportlab fonts.

JGodin-C2C:
This blogpost explains the problem.
http://blog.odoobiz.com/2015/08/odooopenerp-picking-list-printing-blank.html
As reportlab is installed in this image, it should be complete with
batteries.


simahawk:
included into #92

PR #92, opened 2018-12-18
Misc build fixes

JGodin-C2C:


simahawk:
@JGodin-C2C can you check the builds? Both PRs are red

JGodin-C2C:
The builds are still red, but it looks like it is related to python PIP.
Should i try to fix this ? Or you prefer to dig into that ? 
It looks like a probleme in the setup.py ? 
Regards. 
J

JGodin-C2C:
Got the bug . 
fix is on its way

simahawk:
@leemannd would be better to fix the build 1st

leemannd:
At least it's a partial fix.
As you wish.

simahawk:
:fireworks: :tada:  thanks @JGodin-C2C it's all green now (even if does not appear here yet)!

guewen:
Do you know why we never needed the reportlab fonts before and we don't need them on our local containers but need them for a particular customer?

PR #96, opened 2019-01-09
FIX: Set ODOO_DATA_PATH environment variable for Odoo 9.0 and 10.0

sebalix:


10.0/Dockerfile
```diff
@@ -66,6 +66,8 @@ ENV ODOO_VERSION=10.0 \
     DB_PASSWORD=odoo \
     ODOO_BASE_URL=http://localhost:8069 \
     ODOO_REPORT_URL=http://localhost:8069 \
+    # the place where you put the data of your project (csv, ...)
```
simahawk: does it work w/ comments in the middle of the line?

simahawk: well, build is green... seems so...

sebalix: It is done like this for 11.0 and 12.0, so I suppose it works.

PR #98, opened 2019-02-15
set missing locale to fix install of python-ldap

tschanzt:


10.0/Dockerfile
```diff
@@ -7,6 +7,9 @@ RUN mkdir -p /odoo /var/log/odoo
 COPY ./base_requirements.txt /odoo
 COPY ./install /install
 
+ENV LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8
```
guewen: These variables already exist in the ENV at the end of the file (interestingly it was at the same place in https://github.com/camptocamp/docker-odoo-project/blob/d6ce15a5e7306c2fbb4ab2e6fa9b9b071bff1388/11.0/Dockerfile#L14). Which means: you have to remove them from the ENV at the bottom + set a comment here so noone ever will move them to the bottom of the file again :)

tschanzt:
@guewen updated.

tschanzt:
@guewen ping

PR #99, opened 2019-02-15
Unpin pip

tschanzt:


tschanzt:
@leemannd @guewen Updated. I added the history file.

tschanzt:
updated the pr to pin to v 19.0.3 and rebased onto master for fixing tests

yvaucher:
@tschanzt prior to the add of pip in this file we used the default pip version installed by travis.

https://github.com/camptocamp/docker-odoo-project/commit/82b528744aaea5081a815f5b3cd6c96dce00e5a5

Why do you want to pin it?

tschanzt:
@yvaucher if im not mistaken these changes install pip from github so master which we don't want and for v12 i think its good to have pip at the same version for all.

yvaucher:
pip is installed in Dockerfile here

https://github.com/camptocamp/docker-odoo-project/blob/master/12.0/Dockerfile#L40

As pip is the python dependency manager and it is usually a bad idea to manage pip with pip itself I would rather let it unpinned.

I'm in favor of removing pip from requirements.txt

tschanzt:
well the docker file is using pip to reinstall pip :smile: 

We can leave leave it unpinned but its what caused the issue of our python-ldap from 2016 suddendly not building anymore which was fixed with #98 and lead to this pr.
But in the end if you want it unpinned ill do it it seems to more stable now that it once was .

tschanzt:
@yvaucher pinning of pip removed.

PR #104, opened 2019-03-04
Add script to allow setting of mail domain

tschanzt:


simahawk:
@tschanzt why do we need a script? We can use env vars + env ir config param... :thinking:

tschanzt:
@simahawk i think ir_config_param isn't migrated to guewens new server env yet. https://github.com/OCA/server-env/blob/11.0/server_environment_ir_config_parameter/models/ir_config_parameter.py

simahawk:
@tschanzt then do that instead of this hack no? :innocent:

guewen:
@tschanzt it's here: https://github.com/OCA/server-env/pull/15
For ir.config.parameter, there is no change needed regarding my recent changes, it works out of the box.

tschanzt:
closing it since the modifications need to be done elsewhere.

simahawk:
@guewen Timon was pointing out that we cannot use ENVIRON vars, for file env vars we are fine already

guewen:
> Timon was pointing out that we cannot use ENVIRON vars, for file env vars we are fine already

Yes we can, because everything which is in the ENVIRON vars end up in `serv_config`:

https://github.com/OCA/server-env/blob/b6ae38f48eb90eb6bb3b8e578b2ccdae9e2a11ed/server_environment/serv_config.py#L128-L141

Which is what is read from the ir_config_parameter implementation (or other existing implementations)

https://github.com/OCA/server-env/blob/55a0211702a9f60a665800b801e1605376f8ebde/server_environment_ir_config_parameter/models/ir_config_parameter.py#L6

PR #105, opened 2019-03-07
runtests: allow to use a DB template

simahawk:
When running tests locally you might want to simulate tests as on Travis
and at the same time to not start from scratch every time
because your time IS important ;)

You can now run the tests like:

`doco run --rm -e DB_TEST_TEMPLATE=foo odoo runtests`

### DIFF
--- a/bin/runtests
+++ b/bin/runtests
@@ -52,7 +52,16 @@ DB_NAME_TEST=${DB_NAME}_test
 
 
 echo "Create database"
-createdb -O $DB_USER ${DB_NAME_TEST}
+if [[ ! -z "$DB_TEST_TEMPLATE" ]];
+then
+  # When running tests locally you might want to not start from scratch every time.
+  # Tip: create the template w/ only `base` or few core modules installed
+  # to still replicate more or less the same behavior as on travis.
+  echo "using template: ${DB_TEST_TEMPLATE}"
+  createdb -O $DB_USER ${DB_NAME_TEST} -T ${DB_TEST_TEMPLATE}
+else
+  createdb -O $DB_USER ${DB_NAME_TEST}
+fi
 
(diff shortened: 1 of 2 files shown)


bin/runtests
```diff
@@ -52,7 +52,16 @@ DB_NAME_TEST=${DB_NAME}_test
 
 
 echo "Create database"
-createdb -O $DB_USER ${DB_NAME_TEST}
+if [[ ! -z "$DB_TEST_TEMPLATE" ]];
```
guewen: Cool :) can you add a note in the readme?

simahawk: done

bin/runtests
```diff
@@ -52,7 +52,16 @@ DB_NAME_TEST=${DB_NAME}_test
...
+then
+  # When running tests locally you might want to not start from scratch every time.
+  # Tip: create the template w/ only `base` or few core modules installed
+  # to still replicate more or less the same behavior as on travis.
+  echo "using template: ${DB_TEST_TEMPLATE}"
+  createdb -O $DB_USER ${DB_NAME_TEST} -T ${DB_TEST_TEMPLATE}
```
yvaucher: I think that when using a template, you want to skip the step `Install official/OCA modules`

Otherwise, there is a risk that you will try to install modules which are already installed. And this is the step you want to skip.

You can let the remainder of modules you don't have in your database be installed when we install the local-src modules.

simahawk: @yvaucher shall I wrap from line 63 to line 90 into the else block of createdb?

simahawk: @yvaucher ping :)

yvaucher: either we add a variable and test it each time (which is bad to the point of view of the else) or as you said we wrap it.
I think the wrap would be the more readable one and simple to fix.

PR #107, opened 2019-04-15
Bump Jinja version to 2.10.1

yvaucher:
Bump Jinja version

and fix issue with libc-ares2

### DIFF
--- a/install/package_odoo_9.0_10.0.sh
+++ b/install/package_odoo_9.0_10.0.sh
@@ -2,6 +2,8 @@
 set -eo pipefail
 
 apt-get update
+# authentication error for libc-ares2
+apt-get install -y --force-yes --no-install-recommends libc-ares2
 apt-get install -y --no-install-recommends \
     antiword \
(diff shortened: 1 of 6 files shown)


simahawk:
@yvaucher :red_circle:

yvaucher:
@simahawk Yep, on jessie seems the package libc-ares2 is not registered with the proper key.

I tried `apt key-update` but it didn't solve it.
Using `--force-yes` allow us to fix the build, but that seems to be a temporary fix.

install/package_odoo_9.0_10.0.sh
```diff
@@ -2,6 +2,8 @@
 set -eo pipefail
 
 apt-get update
+# authentication error for libc-ares2
```
simahawk: can you open an issue to not forget about this?

yvaucher: https://github.com/camptocamp/docker-odoo-project/issues/108

PR #109, opened 2019-04-24
Bump urllib3 version to 1.24.2

yvaucher:
Bump urllib3 version


yvaucher:
No specific reason, I followed recommendation to fix a CVE which requires 1.24.2 or later.

PR #110, opened 2019-05-03
Add support for legacy projects

p-tombez:
* Add support for v7
* Add support for v8


7.0/bin-compat/anthem
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+# "Hack" to avoid crashing tests on legacy Odoo versions
+
+set -e
+echo "WARNING! Legacy project => Anthem will not do anything"
+exit 0
```
guewen: Do we want to exit 1 so it's a hard failure?

p-tombez: I think the tests on travis might fail, thus not pushing the image to the hub, if we have a hard failure here. Not sure, tho.

guewen: Got it

example/odoo/migration.yml
```diff
@@ -1,6 +1,6 @@
 migration:
   options:
-    install_command: odoo
+    install_command: odoo --without-demo=${WITHOUT_DEMO:-True}
```
Tonow-c2c: that is compatible with Version > 8 ?

example/odoo/Dockerfile
```diff
@@ -16,6 +16,7 @@ COPY ./setup.py /odoo/
 COPY ./VERSION /odoo/
 COPY ./migration.yml /odoo/
 
+RUN replace_dependencies.sh
```
guewen: Shouldn't we put it in the entrypoint?

p-tombez: Too late, it has to be done before the next 2 lines in this file

guewen: Right!

bin/replace_dependencies.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
```
guewen: The fact that we need either to

* add this in the Dockerfile legacy project (7.0, 8.0, not sure which requires it)
* pull OCB after https://github.com/OCA/OCB/commit/b2e48ad8b7cbd62d366e6ffee1861a6085999ce0

Should be documented in the readme (we should have a specific "legacy 7.0/8.0" section in the readme)

bin/runtests
```diff
@@ -72,16 +72,18 @@ else
...
         mkdir -p "$CACHE_DIR"
         pg_dump -Fp -d $DB_NAME_TEST -O -f "$CACHED_DUMP"
     fi
 fi
 echo "🔧🔧 Install local-src modules 🔧🔧"
-odoo --stop-after-init --workers=0 --database $DB_NAME_TEST --log-level=warn --without-demo="" -i ${LOCAL_ADDONS}
-coverage run --source="${LOCAL_SRC_DIR}" "${ODOO_BIN_PATH}" --stop-after-init --workers=0 --database $DB_NAME_TEST --test-enable --log-level=test --log-handler=":INFO" -u ${LOCAL_ADDONS}
+gosu odoo odoo --stop-after-init --workers=0 --database $DB_NAME_TEST --log-level=warn --without-demo="" -i ${LOCAL_ADDONS}
+
+export COVERAGE_FILE=/home/odoo/.coverage
```
guewen: I guess this is a required change in order to run the tests as odoo (won't be able to write in `/.coverage`), can you document it in the changelog (if we have volumes using the path we have to adapt it)?

example/odoo/migration.yml
guewen: @Tonow-c2c yes it is

@p-tombez it triggers some memories to me... `--without-demo` buggy in the config file... something to document in the 'legacy section' of the docs as well :)

PR #113, opened 2019-07-23
Fix missing variable with docker-compose exec

sebastienbeau:
When using docker-compose exec the entrypoint is not executed and so
the export of psql variable is not done.
This change will modify the bashrc configuration so if you do a
docker-compose exec odoo gosu odoo bash
you will have the right psql variable

@guewen @hpar

The alternative solution will be to use psql vars into environment file instead of custom one; but this is a broken change

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -15,6 +15,17 @@ export PGUSER=${DB_USER}
 export PGPASSWORD=${DB_PASSWORD}
 export PGDATABASE=${DB_NAME}
 
+# As docker-compose exec do not launch the entrypoint
+# init PG variable into .bashrc so it will be initialized
+# when doing docker-compose exec odoo gosu odoo bash
+echo "
+export PGHOST=${DB_HOST}
+export PGPORT=${DB_PORT}
+export PGUSER=${DB_USER}
+export PGPASSWORD=${DB_PASSWORD}
+export PGDATABASE=${DB_NAME}
+" >> /home/odoo/.bashrc
+
 # Accepted values for DEMO: True / False
 # Odoo use a reverse boolean for the demo, which is not handy,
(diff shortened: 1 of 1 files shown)


sebastienbeau:
I have added the comment.
Note you need to enter in the container with the user "odoo" with bash
so something like  docker-compose exec odoo gosu odoo bash.

The best will be to use directly the PG vars into the odoo template, but this is a broken change

bin/docker-entrypoint.sh
```diff
@@ -15,6 +15,17 @@ export PGUSER=${DB_USER}
 export PGPASSWORD=${DB_PASSWORD}
 export PGDATABASE=${DB_NAME}
 
+# As docker-compose exec do not launch the entrypoint
```
rvalyi: s/do not/does not/
May be you could also add the comment:
```The best would be to use directly the PG vars into the odoo template, but would be a breaking change```\
so eventually this is refactored better in the future.

hparfr:
>The best will be to use directly the PG vars into the odoo template, but this is a broken change
+1 for this.

guewen:
Functionality is cool, what about security? Usually we use a [`.pgpass`](https://www.postgresql.org/docs/9.3/libpq-pgpass.html) for storing the password, which is `chmod 0600`, but I am not sure of the risks in Docker's context.

hparfr:
From the doc https://www.postgresql.org/docs/current/libpq-envars.html
> PGPASSWORD behaves the same as the password connection parameter. Use of this environment variable is not recommended for security reasons, as some operating systems allow non-root users to see process environment variables via ps; instead consider using a password file (see Section 34.15).

I don't think it will be an issue in our docker context.


@guewen  So you use both .pgpass and $DB_PASSWORD ?

PR #115, opened 2019-08-28
Add script to install wkhtml 0.12.5.1

bealdav:
## ISSUE

Above the result with 0.12.4

Probably broken with last odoo commits.

wkhtml

Below with 0.12.5.1

## CONTENT

- [X]  set wkhtmltopdf new version

### DIFF
--- a/12.0/Dockerfile
+++ b/12.0/Dockerfile
@@ -35,7 +35,7 @@ RUN set -x; \
         && /install/package_odoo_11.0_12.0.sh \
         && /install/setup-pip.sh \
         && /install/postgres.sh \
-        && /install/wkhtml_12_4.sh \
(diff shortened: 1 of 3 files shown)


bealdav:
Please @guewen @sebastienbeau 
Could you make recommandation here ?

guewen:
We had this here https://github.com/camptocamp/docker-odoo-project/pull/62
Can you read the discussion? For versions < 12.0 I'm not sure we want it by default as it might break the reports if the proper odoo commit is not included.

guewen:
For 12.0 of course we should have it

bealdav:
ping @hparfr

guewen:
Thanks!

PR #126, opened 2020-01-09
Fix missing .coveragerc for v11-12-13

simahawk:
Not sure if we should remove the lines (or update them) for v < 11

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -28,6 +28,7 @@ Unreleased
 
 * Pin `setuptools<45` and other dependencies to ensure Python 2 support in Odoo<=10
 * Fix deprecated download links for wkhtmltopdf
+* Include `.coveragerc` for all versions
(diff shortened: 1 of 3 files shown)


guewen:
Shouldn't it be in Dockerfile-batteries too?

simahawk:
> Shouldn't it be in Dockerfile-batteries too?

I thought so, @p-tombez told me it shouldn't be necessary (for us at least).

p-tombez:
@guewen I thought Dockerfile-batteries was added on top of the other Dockerfile. If not, then the change need to be done in both files.

simahawk:
done

simahawk:
@guewen @p-tombez good for you?

simahawk:
@p-tombez I've rebased. Would you mind review + merge to include this in next release? :pray:

PR #130, opened 2020-03-11
disable pip version checks

gurneyalex:


guewen:
Thanks! I wonder if we can put the option in `/etc/pip.cfg` as done here https://github.com/camptocamp/docker-odoo-project/blob/49a2c211f98d760c387f313f9266dbb921d66b86/install/setup-pip.sh

p-tombez:
Thanks, but as suggested by @guewen , it could be better to have that in the configuration file. Plus, as you can see in the failed builds, older pip version does not support this argument

yvaucher:
Seems a good idea and trivial to do.

I created a PR here with the edition in pip.conf:

https://github.com/camptocamp/docker-odoo-project/pull/140

PR #132, opened 2020-03-17
Build fixes (python libraries + wkhtmltopdf)

### LINKED ISSUE #128: Builds failing due to EOL of Python 2
Builds fails with error message
```
    You are running Setuptools on Python 2, which is no longer
    supported and
    >>> SETUPTOOLS WILL STOP WORKING <<<
    in a subsequent release (no sooner than 2020-04-20).
    Please ensure you are installing
    Setuptools using pip 9.x or later or pin to `setuptools<45`
    in your environment.
    If you have done those things and are still encountering
    this message, please comment in
    https://github.com/pypa/setuptools/issues/1458
    about the steps that led to this unsupported combination.
```

p-tombez:
* Pin setuptools and dependencies to ensure Python 2 support in Odoo<=10
* Fix deprecated download links for wkhtmltopdf

Fixes #128

### DIFF
--- a/7.0/Dockerfile
+++ b/7.0/Dockerfile
@@ -39,7 +39,7 @@ RUN set -x; \
         && /install/postgres.sh \
         && /install/wkhtml_12_1_2.sh \
         && /install/dev_package.sh \
-        && pip install -U pip setuptools \
+        && pip install -U pip "setuptools<45" \
         && pip install -r /odoo/base_requirements.txt \
         && /install/purge_dev_package_and_cache.sh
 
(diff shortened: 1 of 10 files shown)


7.0/Dockerfile
```diff
@@ -39,7 +39,7 @@ RUN set -x; \
         && /install/postgres.sh \
         && /install/wkhtml_12_1_2.sh \
         && /install/dev_package.sh \
-        && pip install -U pip setuptools \
+        && pip install -U pip "setuptools<45" \
```
Tonow-c2c: && pip install -U pip setuptools<45 didn't work ?

(just to know)

p-tombez: `<` character is used to redirect input, you have to put the quotes to ignore this behavior. If not, you'll get an error like `no such file or directory: 45`

Tonow-c2c: thanks a lot for this explanation !

PR #136, opened 2020-04-09
Fix PostgreSQL package installation in v13

p-tombez:


install/package_odoo_13.0.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eo pipefail
 
-apt-get update
+apt-get update -o Acquire::AllowInsecureRepositories=true
```
yvaucher, line 4: Shouldn't we add the apt-key each time instead?
We might need to install then uninstall curl for that.

p-tombez: The key is actually added here : https://github.com/camptocamp/docker-odoo-project/blob/master/install/postgres.sh#L4
But the apt-get update occurring here is prior to this addition but after adding the repository. I only had to add this because in buster, the warning we had in all other versions became an error, thus making the build fail.
I didn't want to modify the flow of the installation to keep consistency between all versions

PR #139, opened 2020-05-15
shell and --help bypass entrypoints

yvaucher:
This might already have happened to you. You just want to launch a shell to check something in a database or launch the `--help` to check odoo parameters. Before this PR you would likely have to add explicitly a `-e MIGRATE=False`

The 2 following commands do not trigger entrypoints anymore
and are be executed straight forward.

docker-compose run --rm odoo odoo shell [...]
docker-compose run --rm odoo odoo [...] --help [...]

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -109,6 +109,9 @@ if [ ! "$(stat -c '%U' /var/log/odoo)" = "odoo" ]; then
 fi
 
 BASE_CMD=$(basename $1)
+CMD_ARRAY=($*)
+ARGS=(${CMD_ARRAY[@]:1})
+
 if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] ; then
 
   BEFORE_MIGRATE_ENTRYPOINT_DIR=/before-migrate-entrypoint.d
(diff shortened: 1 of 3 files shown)


bin/docker-entrypoint.sh
```diff
@@ -109,20 +109,26 @@ if [ ! "$(stat -c '%U' /var/log/odoo)" = "odoo" ]; then
...
 
-  if [ -z "$MIGRATE" -o "$MIGRATE" = True ]; then
-    gosu odoo migrate
-  fi
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] ; then
 
-  START_ENTRYPOINT_DIR=/start-entrypoint.d
-  if [ -d "$START_ENTRYPOINT_DIR" ]; then
-    run-parts --verbose "$START_ENTRYPOINT_DIR"
+  # Bypass entrypoints when `odoo shell` or `odoo --help` are used
```
guewen, line 117: While bypassing migrate makes sense, I don't think bypassing entrypoints is a  good idea. Some entrypoints *have* to be run before odoo is executed in dev (disable email server for instance) and so they must be applied for shell as well.

yvaucher: Ok, I'll limit to the migration and keep the entrypoints.

yvaucher:
Updated and documented

PR #143, opened 2020-07-22
Fix dbfilter in test environment

mmequignon:
When in test environments, `dbfilter` still refers to `odoodb` in `/etc/odoo.cfg`.

Problem appears in HttpCase when trying to authenticate, dbfilter denies the connection.

```bash
2020-07-20 00:00:00,000 102 WARNING ? odoo.http: Logged into database 'odoodb_test', but dbfilter rejects it; logging session out.
```

### DIFF
--- a/bin/runtests
+++ b/bin/runtests
@@ -49,6 +49,7 @@ fi
 DEPS_ADDONS=$(list_dependencies.py "$LOCAL_ADDONS")
 
 DB_NAME_TEST=${DB_NAME}_test
+sed -i "s/= ^${DB_NAME}/= ^${DB_NAME_TEST}/g" /etc/odoo.cfg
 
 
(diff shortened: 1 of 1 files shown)


bin/runtests
```diff
@@ -49,6 +49,7 @@ fi
 DEPS_ADDONS=$(list_dependencies.py "$LOCAL_ADDONS")
 
 DB_NAME_TEST=${DB_NAME}_test
+sed -i "s/= ^${DB_NAME}/= ^${DB_NAME_TEST}/g" /etc/odoo.cfg
```
guewen, line 52: As we use `--database $DB_NAME_TEST` in the command lines below in the script, I'd prefer to be consistent and use `--db-filter`

PR #144, opened 2020-07-27
Disable the cache for pip

yvaucher:
This doesn't provide any significant change in size. It delegates the responsibility of not storing his own cache to pip instead of having to sweep everything after.

IMO it makes more sense than a post cleaning.


guewen:
It should already be the case with https://github.com/camptocamp/docker-odoo-project/blob/b5405ff7c63c73470f489fe8645f62ac2d57ed16/install/setup-pip.sh#L9

yvaucher:
Ok thanks @guewen

PR #147, opened 2020-10-15
Fix wkhtmltopdf version for Odoo version <= 10

p-tombez:
This PR is intended to circumvent the deletion of the package for wkhtmltopdf version 0.12.1.2 on Odoo side as it was considered deprecated.
I manually recompiled the package, but the solution to integrate it in the image proposed here is far from being optimal, so I gladly update it if you have any better idea

### DIFF
--- a/install/wkhtml_12_1_2.sh
+++ b/install/wkhtml_12_1_2.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


yvaucher (approved):

yvaucher:
Multi-stage build is likely what you are looking for

https://docs.docker.com/develop/develop-images/multistage-build/#use-multi-stage-builds

PR #150, opened 2020-11-05
Add a 12.0-buster image

yvaucher:


yvaucher:
This currently fails with the following error.


```
Step 8/20 : RUN set -x;         sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" > /etc/apt/sources.list.d/pgdg.list'         && /install/package_odoo_11.0_12.0.sh         && /install/setup-pip.sh         && /install/postgres.sh         && /install/wkhtml_12_5.sh         && /install/dev_package.sh         && python3 -m pip install --force-reinstall pip setuptools         && pip3 install -r /odoo/base_requirements.txt --ignore-installed         && /install/purge_dev_package_and_cache.sh
 ---> Running in 99c6cb4277b2
+ sh -c echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+ /install/package_odoo_11.0_12.0.sh
Get:1 http://deb.debian.org/debian buster InRelease [121 kB]
Get:2 http://security.debian.org/debian-security buster/updates InRelease [65.4 kB]
Get:3 http://apt.postgresql.org/pub/repos/apt stretch-pgdg InRelease [76.6 kB]
Get:4 http://deb.debian.org/debian buster-updates InRelease [51.9 kB]
Get:5 http://security.debian.org/debian-security buster/updates/main amd64 Packages [244 kB]
Get:6 http://deb.debian.org/debian buster/main amd64 Packages [7906 kB]
Err:3 http://apt.postgresql.org/pub/repos/apt stretch-pgdg InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 7FCC7D46ACCC4CF8
Get:7 http://deb.debian.org/debian buster-updates/main amd64 Packages [7856 B]
Reading package lists...
W: GPG error: http://apt.postgresql.org/pub/repos/apt stretch-pgdg InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 7FCC7D46ACCC4CF8
E: The repository 'http://apt.postgresql.org/pub/repos/apt stretch-pgdg InRelease' is not signed.
The command '/bin/sh -c set -x;         sh -c 'echo "deb http://apt.postgresql.org/pub/repos/apt/ stretch-pgdg main" > /etc/apt/sources.list.d/pgdg.list'         && /install/package_odoo_11.0_12.0.sh         && /install/setup-pip.sh         && /install/postgres.sh         && /install/wkhtml_12_5.sh         && /install/dev_package.sh         && python3 -m pip install --force-reinstall pip setuptools         && pip3 install -r /odoo/base_requirements.txt --ignore-installed         && /install/purge_dev_package_and_cache.sh' returned a non-zero code: 100
+ on_exit
+ echo 'Cleaning up temporary directory...'
Cleaning up temporary directory...
+ rm -rf /tmp/tmp.JiCp2DGnxx
+ rm -f /tmp/odoo.tar.gz
Makefile:16: recipe for target 'build' failed
make: *** [build] Error 100
```

rvalyi:
thanks a lot we will start testing on our side.

12.0-buster/Dockerfile
```diff
@@ -0,0 +1,78 @@
...
+# Moved because there was a bug while installing `odoo-autodiscover`. There is
+# an accent in the contributor name
+ENV LANG=C.UTF-8 \
+    LC_ALL=C.UTF-8
+
+# build and dev packages
+ENV BUILD_PACKAGE \
+    build-essential \
+    gcc \
+    python3.5-dev \
```
rvalyi, line 19:
```suggestion
    python3.7-dev \
```

rvalyi:
Hello @yvaucher  thank you for your work,

Postgres client install was successful but python3.5-dev package install failed with:
```
Reading package lists...
Reading package lists...
Building dependency tree...
Reading state information...
E: Unable to locate package python3.5-dev
E: Couldn't find any package by glob 'python3.5-dev'
E: Couldn't find any package by regex 'python3.5-dev'
```

could you try the above suggestion to change it for `python3.7-dev` ?

yvaucher:
@rvalyi thanks
it worked

I pushed the following tag already:

```
camptocamp/odoo-project:12.0-buster-latest
camptocamp/odoo-project:12.0-buster-latest-onbuild
camptocamp/odoo-project:12.0-buster-latest-batteries
camptocamp/odoo-project:12.0-buster-latest-batteries-onbuild
```

rvalyi:
Thank you @yvaucher we will try and give feedback. Another very easy thing you guys can do at C2C is take your favorite OCA repos and give it a 3.7 test run with trivial PRs such as https://github.com/OCA/l10n-brazil/pull/1021/files. BTW would be cool if the OCA move on and start testing on say Python 3.6 instead of EOL 3.5 on v12 repos.

bealdav:
I think that always use the lowest release possible python version will always lead us to this kind of problem (aka follow strict odoo version policy).

test too many python versions lead us to deep earth environmental impact see https://github.com/OCA/oca-addons-repo-template/pull/29#issuecomment-714674787

test too many python versions lead us to a cost instead of free price at mid term (CI tools like travis change their policy over the time)

Isn't the time to choose a LTS like python version instead of the oldest one ? In oca and here ?

yvaucher:
@bealdav that's a talk to have with OCA. I agree you don't want to test all version and make your tests with what you want to guarantee that works.

Firstly we should acknowledge that most installation are done with the default python version on Ubuntu or debian.
Then, based on that do we want to encourage going on a different version. Or do we want to ensure our modules works with the default?

IMO the most up to date system is better.
Is LTS python version a thing?
Do you mean we want to avoid end-of-life versions?

yvaucher:
@rvalyi I haven't tested it, and one thing to test is wkhtmltopdf, not sure it works as expected.

bealdav:
> Do you mean we want to avoid end-of-life versions?

exactly

bealdav:
I already open the discussion

https://odoo-community.org/groups/contributors-15/contributors-176724

It's time to have C2C feedback according to environment and cost problematic

guewen:
I already answered on that thread and the fact that we don't have much choice: if odoo supports 3.5, we have to ensure our modules work this 3.5. Of course, I would love to use new 3.7 features, but what a mess if odoo supports 3.5 and the modules support for python versions is fragmented. As for testing, then it makes sense to test with the minimal version as it will support upwards versions.

yvaucher:
As stated by Olivier Dony in the thread you mention, is that as long as an end-of-life python version is distributed it causes no issue.
In reality it mainly means that we have to pin the python packages version to old versions are they will slowly drop support of end-of-life python version.

The best way to get the more up to date system might be to also update the Odoo version, but as we know it still a pain to do. What block us on this is the migration complexity. And the upgrade tools that needs love.

bealdav:
As said by O. Dony https://github.com/odoo/runbot/issues/414#issuecomment-710311902

> On the Odoo Runbot side, the goal is different. **We should ideally test all supported Python versions for each series**. But we don't have the infrastructure for that now, so we have to settle for an acceptable common ground, which is currently Python 3.6 (it works from v11 to v14).

That's a big problem: 'ideally' is probably not the better word for the planet or reduce implied cost/maintenance, for OCA and partners.

Are you OK @yvaucher  to say that we should invite Odoo SA for future versions to switch to a better policy for the planet and maintainability ?

rvalyi:
Hello, now that the thread fully is hijacked, taking the opportunity to say I agree wit @bealdav in the sense that Odoo SA do programmed obsolescence of its versions to strongly incentive companies to migrate using their services. Or at best they don't care about their real customer deployments once they have more new ones. We all know that supporting an ERP version only 3 years is a joke considering:

1. how hard/expensive it is to migrate
2. considering that using Odoo EE or OpenUpgrade, you can only migrate to version N when N+1 is out, meaning following Odoo SA theory you just spent like 30% of your ERP project migrating and you can only enjoy it for 1 years before investing the same 30% to migrate again to avoid being on an unsupported version. Example you did it perfect and migrated from 10 to 12 in 2019, in 2020 you should be investing a lot again to be on v14 in 2021 and you will still be on an unsupported version like 6 months.

This is just not the real world, both C2C and us know that perfectly. So the argument: it's because Odoo SA position is blablabla is just not serious. A more mature attitude from the OCA could be: look, v14 is out, chances are you will still be using it in 2024, forget about Python 3.6 now and let's use Python 3.7 or 3.8 already. (and like David I think testing multiple Python versions is really not environment friendly considering how long take the tests). My 2 cts.

guewen:
>  in the sense that Odoo SA do programmed obsolescence of its versions to strongly incentive companies to migrate using their services

How does it relate to the python version to test? I would argue that supporting older python versions is the opposite of programmed obsolescence...
Honestly, I don't understand the discussion. Testing only one version, I get the point, but choosing new versions (3.7+) to test code that must be compatible with 3.5+ I don't get it. Using a recent version for runtime, I'm all in, but not for tests (which is point raised by @bealdav).

guewen:
Anyway, you should have this discussion with Odoo then, because as long as they support 3.5 or 3.6, we have to as well.

rvalyi:
> > in the sense that Odoo SA do programmed obsolescence of its versions to strongly incentive companies to migrate using their services
> 
> How does it relate to the python version to test?

Because when on v14 for instance the OCA says all right dudes we support 3.6 even if it's EOL next year, what happens in that in 2 or 3 years lots of v14 deployments will have obsolescence issues: they will be in unsuported/unmaintained distros with loads of open CVEs (how cool is that when you have an Odoo ecommerce?) and have a lot more work doing any development to find out which last version of package X will still work on 3.6, how bad is that with the lack of bugfix or security fixes in these old packages (think 3.6 in 3 years ). We all know fixing things before producting is way cheaper than once everything is doing missing critical work.

A lot of this could be avoided instead if the OCA bite the bullet and assume it's useless to support version 3.6 on 14 or keep testing v12 in say 2021 or 2022 while only a fool would be running Odoo 12.0 on 3.5 by that time.

yvaucher:
As @guewen pointed it out. Here it seems there is some mixing between

**compatibility**

We want to be compatible with as many version as possible (the range supported by the core) to not force anyone to change their python version. For OCA modules you want to test to the lowest as backward compatibility is available.

**Runtime**

We want to have last performance improvements from python and thus run Odoo on most recent versions. And here docker-odoo-project is currently lacking this.

Having both in mind, it means that we can gain from last python version improvements but not for the breaking changes in the code. Like the f strings which are a nice feature but would break backward compatibility and thus increasing obsolescence.

guewen:
> Because when on v14 for instance the OCA says all right dudes we support 3.6 even if it's EOL next year, what happens in that in 2 or 3 years lots of v14 deployments will have obsolescence issues

How is it worse than having odoo supporting 3.6 and people having crashing OCA modules because they do not support the same version. Sorry, but that's nonsense. And if people still decide to use 3.6, just because they can, that's on them.

yvaucher:
@rvalyi It's unlikely that python package distribution will disappear soon. If you want to do python2 you can still grab the old package version. You won't get new features nor security fixes, but you can still build a stable version.

End-of-life means that package can drop support in their new releases. But it doesn't mean they will erase existing ones.

You are right that for maintenance it will be stopped. This is an issue for the runtime, CVE are package specifics and might not affect all projects. Maybe on the long run some CVE will affect all existing Odoo project using python 3.5. However giving the possibility to use python 3.5 doesn't force to use python 3.5.

To me this is not the responsibility of OCA.

Nevertheless, the reflexion is enlightening and we should consider having more up to date version of python in docker-odoo-project.

rvalyi:
> > Because when on v14 for instance the OCA says all right dudes we support 3.6 even if it's EOL next year, what happens in that in 2 or 3 years lots of v14 deployments will have obsolescence issues
> 
> How is it not worse than having odoo supporting 3.6 and people having crashing OCA modules because they do not support the same version? Sorry, but that's nonsense. And if people still decide to use 3.6, just because they can, that's on them.

In a word, you are forcing the future +3 years ecosystem (say v14) to be a lot more obsolete than it has to, just because you want to please people who are fool enough to go Python 3.6 on their new Odoo 2020 project while it is EOL next year. It's not because Odoo SA has this kind of unresponsible sale behavior than we the OCA have too... And also Odoo SA only care about 350 modules for a demo ERP that is way easier to migrate than real life Odoo deployements like we all do.

The problem is not so much about package disapearing, the problem is much more about the OCA forcing you to use Python 3.5 packages on v12 next year or v3.6 packages on v14 in 3 years. Because yes these packages will be abandoned, with bugs and serious security issues. Or often newer solutions are beating the old legacy packages you will be forced to use. We cannot on one hand, go with Fabien idea of web enabled ERP or even ecommerce enable ERP and on the other hand force real life installations to use obsolete Python packages very early when this could have been avoided with a bit more real life talking from the OCA.

bealdav:
Ok guys:
- some of us (in whole oca) want to take care of old python versions compatibility: we can deal with that up to v14
- we have to take of EOL up to v14 on this docker img : thanks @yvaucher for this PR
- we can try to make odoo sa ask itself on this topic for v15

Thanks to have share your opinions

rvalyi:
@bealdav Given Odoo SA interrest, I'm pretty sure Fabien's stance on this one will be at the same level of its answer to "if you love the OCA so much why is there no link to the OCA on odoo.com??"

@guewen you seems to consider 3.6 incompatibilies in 2020 are equal to say v14 Python 3.6 package obsolescence say in 2023+. To which I say no because the guy going 3.6 on v14 now is just a fool. While people who will still be running v14 in 2023 are very serious people, the bests in fact as migrating large projects every 2 years is just not real as I explained. So that's accepting noobs put a huge burden on people really doing the hard work at the OCA. IMHO not OCA job to lie like Odoo SA about what it really takes to succeed in your Odoo projects.

@ all. As you saw, @bealdav brought this topic to the OCA. It's just that we have not be heard and think we can still try to convince people individually when given the opportunity like here.

CVEs and bugs don't affect only the Python runtime but also the packages still compatible with obsolete runtimes. It can be website HTML or XSS escaping packages, it can be connector packages. We have very real examples with SOAP support for electronic invoicing for instance.

For very long, because Odoo was so slow to embrace Python 3 we had simply no choice. But now that this jump has been made, and now that Odoo has evolved to be much more exposed to the world and interconnected I think this issue is hotter than ever.

For me reasonable stance would be v14 to move on requiring 3.7 say next year and 12 requiring 3.6. A wise balance between when it's reaonsable to ask people to upgrade their distro and not holding hostage the whole ecosystem with obsolete things.

bealdav:
https://youtu.be/Iwb6u1Jo1Mc?t=63

sebastienbeau (approved):
@yvaucher @guewen 
I have tested the image and it's work correctly
Thanks for your work

rvalyi (approved):
This image is well tested on our side now. Also not sure if you have seen but the Travis issues have nothing to do with this image (it's the images before 11.0 which are failing). So I think you can merge...

bealdav:
Thanks a lot

PR #154, opened 2021-04-07
odoo-autodiscover is not necessary for Odoo 11.0 and later

yvaucher:
According to https://github.com/acsone/odoo-autodiscover#odoo-autodiscover

Thus we should be able to safely remove it and restore the CI


yvaucher:
> 12.0 error on autodiscovery passed

PR #155, opened 2021-04-07
[<=10.0] Fix legacy build for Py2 based versions

yvaucher:
With the depreciation of Python 2 some libraries have dropped compatibility.

This restores the build for legacy version (and hopefully make the CI greener)


yvaucher:
All green (finally)

PR #157, opened 2021-04-07
[11.0,12.0] Pin watchdog before Py3.5 compat break

yvaucher:
All version after watchdog 1.0.0 dropped support for Py3.5


yvaucher:
11.0 and 12.0 build fixed

PR #160, opened 2021-06-25
Allow to pass DB_FILTER env var

PierrickBrun:
This PR allows to use DB_FILTER to customize what databases are available in the server.

This can be very useful in a development context with `DB_FILTER = .*`


rvalyi (approved):

10.0/templates/odoo.cfg.tmpl
```diff
@@ -6,7 +6,7 @@ db_host = {{ .Env.DB_HOST }}
 db_name = {{ .Env.DB_NAME }}
 db_user = {{ .Env.DB_USER }}
 db_password = {{ .Env.DB_PASSWORD }}
-dbfilter = ^{{ default .Env.DB_NAME "odoodb" }}$
+dbfilter = ^{{ default .Env.DB_FILTER .Env.DB_NAME }}$
```
simahawk, line 9: shouldn't you keep `"odoodb"` as default?

PierrickBrun: Well `db_name` has no default so I expect it to be always populated.

Also I don't see any reason to have `db_name` not included in `dbfilter` (for instance if DB_NAME='db' and DB_FILTER not set)

simahawk: yeah, you are right. We always have DB_NAME anyway.

simahawk (approved):
@p-tombez or @yvaucher fine for you?

yvaucher (approved):

PR #90, opened 2018-11-09
Ensure executable bit on script directories

pniederlag:
Fixes #89

PR #162, opened 2021-08-25
Executable bit set on script folders

### LINKED ISSUE #89: missing executable bit in /start-entrypoint.d/001_set_report_url
I noticed that in camptocamp/odoo-project:11.0-latest the /start-entrypoint.d/001_set_report_url is missing the executable bit so it won't be executed.

    docker run --rm --entrypoint "/bin/bash" -ti  camptocamp/odoo-project:11.0-latest
    root@85c327e2a9ee:/# ls -l /start-entrypoint.d/
    total 8
    -rwxrwxr-x 1 root root 991 Oct 31 16:22 000_set_base_url
    -rw-rw-r-- 1 root root 648 Oct 31 16:22 001_set_report_url

szabolcsmaj:
Based on @pniederlag 's PR (GH-90) that fixes #89.

He only missed a parameter from `chmod`. If it's easier to fix and merge that then do it. My goal is to make it work.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -30,7 +30,8 @@ on_exit() {
 }
 
 trap on_exit EXIT
-
+# as executable bit will be inherited to the docker image we explicitly set executable bit on scripts
+chmod -R +x start-entrypoint.d/ before-migrate-entrypoint.d/
 cp -r ${VERSION}/. ${TMP}/
 cp -r bin/ ${TMP}
 cp -rT common/ ${TMP}
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-NAME=sloby/odoo-project
+NAME=camptocamp/odoo-project
 ifndef VERSION
 $(error VERSION is not set)
 endif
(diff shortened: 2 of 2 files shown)


build.sh
```diff
@@ -30,7 +30,8 @@ on_exit() {
 }
 
 trap on_exit EXIT
-
+# as executable bit will be inherited to the docker image we explicitly set executable bit on scripts
+chmod -R +x start-entrypoint.d/ before-migrate-entrypoint.d/
```
i-vyshnevska, line 34: I am not sure this change is needed, maybe @p-tombez give a point

p-tombez (commented):
I don't see the utility of this change for several reasons : 
1. Assuming one of the files in the directories doesn't have the execute bit on, the script will change it and git will detect it, thus making the local repository in a stale state (so the local repo is not anymore a source of truth)
2. If anyone make a PR to add a script and forget to add `+x` on it, the Travis build will fail and that person would have to fix it before it is merged
3. These images are intended to be inherited before being used. This fix will not prevent a failure if anyone add a script in these folders without `+x`

PR #164, opened 2021-09-14
Marabunta update to version 0.10.6

leemannd:


10.0/base_requirements.txt
```diff
@@ -63,7 +63,7 @@ urllib3==1.24.2
 
 # Migration tools
 ruamel.yaml==0.16.13
-marabunta==0.10.5
+marabunta==0.10.6
```
leemannd, line 66: Not available as we need to update the way we are creating new verions of marabunta

leemannd:
This is currently stalled as the newest version of marabunta is currently not available

PR #176, opened 2021-12-06
Fix qrcode generation in 14 (reportlab update)

TDu:
When printing an invoice with a qrcode.
The call to reportlab.graphics.barcode.createBarcodeDrawing
would raise the following error:

    AttributeError("'Image' object has no attribute 'fromstring'")

### DIFF
--- a/14.0/base_requirements.txt
+++ b/14.0/base_requirements.txt
@@ -47,7 +47,7 @@ python-dateutil==2.7.3
 pytz==2019.1
 pyusb==1.0.2
 qrcode==6.1
-reportlab==3.5.13; python_version < '3.8'
+reportlab==3.5.54; python_version < '3.8' # official 3.5.13
 # reportlab==3.5.55; python_version >= '3.8'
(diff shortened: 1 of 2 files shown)


14.0/base_requirements.txt
```diff
@@ -47,7 +47,7 @@ python-dateutil==2.7.3
 pytz==2019.1
 pyusb==1.0.2
 qrcode==6.1
-reportlab==3.5.13; python_version < '3.8'
+reportlab==3.5.54; python_version < '3.8'
```
florentx, line 50: Usage is to keep a reference to the official dep version in a comment:
```suggestion
reportlab==3.5.54; python_version < '3.8'  # official 3.5.13
```

florentx (changes requested):
It needs an entry in `./HISTORY.rst` **Libraries**

14.0/base_requirements.txt
TDu: thanks for your review I updated the pr

PR #177, opened 2022-01-13
Add copy of extra_requirements in Dockerfile

smussie:
[BSRD-498](https://jira.camptocamp.com/browse/BSRD-498)


yvaucher:
extra_requirements.txt file is there for their for the `batteries` flavor. We don't want it in the main image.

It is used here: https://github.com/camptocamp/docker-odoo-project/blob/master/common/Dockerfile-batteries

PR #179, opened 2022-02-15
fix: add migrate for lab in before entry point

vrenaville:


bin/docker-entrypoint.sh
```diff
@@ -131,12 +131,14 @@ BASE_CMD=$(basename $1)
 CMD_ARRAY=($*)
 ARGS=(${CMD_ARRAY[@]:1})
 
-if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] ; then
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || ([ "$BASE_CMD" = "gosu" ] && [ "$CMD_ARRAY" = "odoo migrate" ] ); then
```
p-tombez, line 134:
```suggestion
if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || ([ "$BASE_CMD" = "gosu" ] && [ "${ARGS[@]}" = "odoo migrate" ] ); then
```

p-tombez (changes requested):

PR #181, opened 2022-02-28
bugfix(15.0 docker): restore wkhtmltopdf on python3

gurneyalex:
This restores the compatibility mode of the wkhtmltopdf client with python3-based environments. Without it, it is not possible to print on odoo15 instances.

### DIFF
--- a/15.0/Dockerfile
+++ b/15.0/Dockerfile
@@ -36,6 +36,7 @@ RUN set -x; \
         && /install/setup-pip.sh \
         && /install/postgres.sh \
         && /install/kwkhtml_client.sh \
+        && /install/kwkhtml_client_force_python3.sh \
         && /install/dev_package.sh \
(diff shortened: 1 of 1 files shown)


15.0/Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN set -x; \
         && /install/setup-pip.sh \
         && /install/postgres.sh \
         && /install/kwkhtml_client.sh \
+        && /install/kwkhtml_client_force_python3.sh \
```
mmequignon, line 39: according to this  https://github.com/camptocamp/docker-odoo-project/commit/75992d08b7bc64debf52192d78ccc001139aef76#diff-379d1b6436dbc1f0267c62cfb7463cd321ca5b543c3cd41c6ec56a6008455340L39
We should set an `KWKHTMLTOPDF_SERVER_URL` which I don't find (I probably missed it though).
Have you tried this?

PR #198, opened 2022-06-19
imp: su-exec from gosu (for amd64 on arm emulation)

dnplkndll:
this is for reference on using on M1 style arm to amd64 and was the only change needed. I see there may be a some thing that are not in sync since the fork. 

kwkhtml_client
postgresql-client-14


vrenaville:
@dnplkndll 

Hello,

Thanks for the PR, I'm currently check if remove this (gosu) and running the container with Odoo
I still need to test it in real life and fixing test but it's currently build locally.

https://github.com/camptocamp/docker-odoo-project/pull/213


Vincent

PR #229, opened 2022-11-28
feat: switch to github action

vrenaville:


.github/workflows/build.yaml
```diff
@@ -0,0 +1,200 @@
...
+          cp -r start-entrypoint.d/ ${{ matrix.odoo_serie }}_dir
+          cp -r before-migrate-entrypoint.d/ ${{ matrix.odoo_serie }}_dir
+
+
+      - name: Build
+        uses: docker/build-push-action@v3
+        with:
+          context: ./${{ matrix.odoo_serie }}_dir
+          push: ${{ github.event_name != 'pull_request' }}
+          tags: ${{ steps.meta.outputs.tags }}
```
yvaucher, line 198: Instead of building it a second time, it would be more efficient to download the image we pushed into artifacts.

ivantodorovich: Or use a single job that builds, tests and push in multiple steps.
It would be even faster as it would run on the same vm

yvaucher: Indeed, AFAIK this was copied from an other github action where the middle phase is done in parallel for 2 testing vm.

PR #239, opened 2023-02-21
[16.0] Fix: xlrd version

smussie:
Context: the xlrd version 2.0.1 fails in odoo version 16, resulting in the error seen in the photo bellow. Resetting it back to the way odoo used it or using **xlrd >= 1.0.0** works, I tested in both cases.

adiant_xlrd_version_fail

### DIFF
--- a/16.0/base_requirements.txt
+++ b/16.0/base_requirements.txt
@@ -37,7 +37,8 @@ vobject==0.9.6.1
 Werkzeug==0.16.1
 XlsxWriter==3.0.3
 xlwt==1.3.0
-xlrd==2.0.1
+xlrd==1.1.0; python_version < '3.8'
+xlrd==1.2.0; python_version >= '3.8'
 pyOpenSSL==22.1.0
 
 setuptools<58
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -27,6 +27,8 @@ Unreleased
 
 **Bugfixes**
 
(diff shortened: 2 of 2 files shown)


leemannd (approved):

leemannd:
It comes from  https://github.com/odoo/odoo/blob/16.0/requirements.txt#L46

cyrilmanuel (approved):

vrenaville (changes requested):
You pointing on master, this branch is for the 5.X.X release, for the new images 

Please point your PR to release.4.5.X branch

leemannd:
@vrenaville quick question, shouldn't it be fixed also in new images?

sebalix (commented):
Can you remove your fixup commit about pyOpenSSL and cryptography?
So we could merge only the fix of xlrd.

smussie:
> Can you remove your fixup commit about pyOpenSSL and cryptography? So we could merge only the fix of xlrd.

With great pleasure, because the pyOpenSSL and cryptograhy versions were giving me headache :)

sebalix (approved):

sebalix:
@vrenaville All good for you to merge? Will it possible to create a new release you think even with the build issue?

vrenaville:
@sebalix I'm ok if we are unable to fix it, in a short time. but if we need to downgrade xlrd, I will make a PR on master.

vrenaville (approved):

PR #245, opened 2023-05-09
[12.0-15.0] Pin reportlab for all Python versions

florentx:
By mistake, Reportlab was only pinned with Python >= 3.8

Consequence seen when build tried to install ReportLab 4.0.0 and failed building pycairo binary wheels.


florentx:
wrong base branch

PR #247, opened 2023-05-15
[IMP] include click-odoo-contrib in dependencies

gurneyalex:
This will allow using click-odoo-update in migration scripts

see https://github.com/acsone/click-odoo-contrib


simahawk (approved):

ivantodorovich (approved):
`1.17.0` was released last week, we could pin it instead

https://github.com/acsone/click-odoo-contrib/releases

PR #251, opened 2023-09-07
Give a dedicated cachedb dir for volume mount

yvaucher:
In order to extract the DB dump from the image we want this folder to be standalone especially in the CI

### DIFF
--- a/bin/runmigration
+++ b/bin/runmigration
@@ -35,7 +35,7 @@
 set -e
 
 wait_postgres.sh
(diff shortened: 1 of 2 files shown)


ivantodorovich (approved):
Makes sense

yvaucher:
I'll take the same change I did in https://github.com/camptocamp/docker-odoo-project/pull/252

To be able to overwrite this variable.

ivantodorovich (approved):

PR #252, opened 2023-09-07
Refactor DB cache management in runtests and runmigration

yvaucher:
Gives most variables a default value making them changable with `var=${var:=default}`

Extract db restore and dump in a separate db.sh file.

Change (default) value for cachedb directory to make it easier to set a volume on it.

Add `|| true` in runmigration when doing `ls -v` to ignore the error status and set an empty `CACHED_DUMP` variable.

Improve slightly the messages.

### DIFF
--- a/bin/db.sh
+++ b/bin/db.sh
@@ -1,0 +1,27 @@
+#!/bin/bash
+
+db_restore() {
+    DUMP=$1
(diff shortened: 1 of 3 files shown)


florentx:
Superseded by #299, merged in 2024

PR #255, opened 2023-09-12
[REM] Replace wait_postgres.sh with pg_isready

ivantodorovich:
Just to keep things simpler, we can use [`pg_isready`](https://www.postgresql.org/docs/current/app-pg-isready.html) directly


yvaucher:
Currently failing, is it because it takes more than 30 seconds in our unit tests? weird

ivantodorovich:
I'm not sure. `make VERSION=15.0 test` locally works fine, and the exact same thing fails on travis.
I'll switch to draft for now and try to have a look in detail when I have some time

yvaucher:
We just replaced travis by github, can you rebase to give it a new try?

ivantodorovich:
> We just replaced travis by github actions, can you rebase to give it a new try?

Yay! 🥳 it's ok now

yvaucher:
@ivantodorovich I think we can fully remove dockerize as we were installing it only on this purpose.

Check https://github.com/camptocamp/docker-odoo-project/commit/e900a06a63ef73f6bbc30398e29b813c0a199351 and 
https://github.com/camptocamp/docker-odoo-project/commit/9ddc6685aee44a86fde1a88eb3786894b5c3f7b7

I had a doubt on pg_isready, but it seems that tool was not existing at the time. We tried first an other aproach with was based on `psql` command with was causing issues.

To me once we removed `dockerize` we can merge this.

yvaucher:
I had some doubts about pg_isready but seems we never attempted it, our first try long time ago was with a `psql` query that was poorly doing the job.

PR #256, opened 2023-09-12
[REM] bin-py2: unused python2 binaries

ivantodorovich:
It was originally implemented in fecb51f51c2821ad10a1e34e47ae6727ac9d8ceb, to support the python2 odoo images.

But since support for Odoo <= 10 has been dropped, it's no longer used.


simahawk (approved):
seems legit

yvaucher (approved):
Indeed this can be removed, the same script for python3 exists in `bin`

PR #258, opened 2023-10-03
[IMP] Modified entrypoint to add debugpy capabilities

wouitmil:
With this modification you can use env variable to enable [debugpy](https://github.com/microsoft/debugpy) on a odoo container

DEBUGPY_ENABLE=1
DEBUGPY_PORT=[8888]

Once done, you just have to start you docker with -p 8888:8888 and use [vscode debugger](https://code.visualstudio.com/docs/python/debugging) ([quick demo](https://www.youtube.com/watch?v=ywfsLKRLmf4)) to connect the instance and start debuging

To test it in vscode, you'll need:

1. An entry in your launch.json file. Here is an example:

```json

{
    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
    "version": "0.2.0",
    "configurations": [
        
        {
            "name": "Python Debugger: Remote Attach",
            "type": "debugpy",
            "request": "attach",
            "connect": {
                "host": "localhost",
                "port": 8888
            },
            "pathMappings": [
                {
                    "localRoot": "${workspaceFolder}/odoo",
                    "remoteRoot": "/odoo"
                }
            ]
        },      
    ]
}
```

2. launch your project with the DEBUGPY_ENABLE variable set to 1
```bash
docker-compose -p bsuni-test-image  run -p 8069:8069 -p 8072:8072 -p 8888:8888 -e DEBUGPY_ENABLE=1 --rm --name=odoo odoo --workers=0 
```
3. Debug ;-)

output

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -114,6 +114,18 @@ if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ]; then
     run-parts --verbose "$START_ENTRYPOINT_DIR"
   fi
 
+  if [ ! -z "$DEBUGPY_ENABLE" ]; then
+    if [ -z "$DEBUGPY_PORT" ]; then
+      DEBUGPY_PORT=8888
+    fi
+    env python3 -m pip install debugpy pydevd-odoo
+    # TODO maybe we should set: GEVENT_SUPPORT=True
+    env python3 -m debugpy --listen 0.0.0.0:${DEBUGPY_PORT}  /usr/local/bin/odoo ${ARGS}
+  else
+    exec "$@"
+  fi
+  
+
   exec "$@"
 fi
 


bin/docker-entrypoint.sh
```diff
@@ -114,12 +114,12 @@ if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ]; then
...
   fi
 
-  if [ ! -z "$DEBUGPY_ENABLE" ]; then
+  if [ ! -z "$DEBUGPY_ENABLE" -o "$DEBUGPY_ENABLE" = True ]; then
     if [ -z "$DEBUGPY_PORT" ]; then
       DEBUGPY_PORT=8888
     fi
-    pip3 install debugpy
-    /usr/bin/python3 -m debugpy --listen 0.0.0.0:${DEBUGPY_PORT}  /usr/local/bin/odoo
+    env python3 -m pip install debugpy
```
ivantodorovich, line 121:
```suggestion
    env python3 -m pip install debugpy pydevd-odoo
```

I recommend installing also `pydevd-odoo`, which makes the debugging experience a little better

wouitmil:
> I'm already using `debugpy` with our Odoo images with the following command :
> 
> ```shell
> docker-compose run --name odoo -p 8069:8069 -p 5678:5678 odoo gosu odoo python3 -m debugpy --wait-for-client --listen 0.0.0.0:5678 /usr/local/bin/odoo --workers=0
> ```
> 
> and by adding the required python dependencies in the file `odoo/dev_requirements.txt` (also used to add pdbpp or any required library for dev/debug purposes).
> 
> So I'm not sure that we need to add all this in our image since it already works without it

Yes, it's exactly what I'm doing now with at this moment. Just that it would be easier for devs I think

p-tombez:
Since we use that image not only in development but also on production environment, IMO, we shouldn't put any development/debug tools on it and keep them at the developer level.
Developer tools widens the surface attack on the production environment and also could have the potential for performance impact in some cases

yvaucher:
As stated by @p-tombez this image is also for prod usage, dev tools shouldn't be part of it.

PR #260, opened 2023-10-30
feat: add odoo 17.0

vrenaville:


17.0/base_requirements.txt
```diff
@@ -0,0 +1,85 @@
+# Odoo dependencies
```
ivantodorovich, line 1: out of curiosity, how do you decide which package version to freeze ?

ivantodorovich (approved):

17.0/extra_requirements.txt
```diff
@@ -0,0 +1,45 @@
+# Extra python dependencies
+algoliasearch==2.6.2
+Adyen==7.1.1
+cachetools==5.2.0
+cerberus==1.3.4
+boto3==1.26.7
+factur-x==2.3
+invoice2data==0.3.6
+mailjet-rest==1.3.4
+openupgradelib==3.3.4
```
leemannd, line 10:
```suggestion
openupgradelib==3.5.0
```

bin/docker-entrypoint.sh
```diff
@@ -86,7 +86,7 @@ if [ ! -f "${CONFIG_TARGET}" ]; then
 fi
 
 # Wait until postgres is up
-pg_isready
```
ivantodorovich, line 89: Hm, sorry about this. I proposed the use of `pg_isready` (https://github.com/camptocamp/docker-odoo-project/pull/255), but after seeing this red build I realize it's not working as expected.

It seems `pg_isready` is not enough: https://stackoverflow.com/questions/35069027/docker-wait-for-postgresql-to-be-running

There's a 100% dockerized solution to this, but it requires changes in the docker compose files, basically:

```yaml
  db:
    ...
    healthcheck:
      test: [ "CMD", "pg_isready", "-q", "-h", "localhost" ]
      interval: 5s
      timeout: 5s
      retries: 5
  odoo:
    ...
    depends_on:
      db:
        condition: service_healthy

```

NOTE: `-h localhost` seems to be very important when running `pg_isready` in the db container, as it otherwise may lead to false positives due to the db that's started in the init process.

With this solution we wouldn't really need to wait in the odoo container like we do now, as everything would be handled by docker. However, it's not such a bad idea to wait here too.. but, unfortunately, `pg_isready` is not enough 😢 

We'd need a wrapper like this it seems:

```bash
#!/bin/bash
set -e

# Default value (set to 0 to disable)
PGTIMEOUT=${PGTIMEOUT:-10}

if [ ${PGTIMEOUT:-10} -le 0 ]; then
    exit 0
fi

if [ -z "$PGHOST" ]; then
    echo "ERROR: PGHOST is not defined" 1>&2
    exit 1
fi

echo -n "Waiting for PostgreSQL to be ready ..."
start_time=$(date +%s)
until pg_isready -q -t 1 >/dev/null; do
    current_time=$(date +%s)
    elapsed_time=$((current_time - start_time))
    if [ "$elapsed_time" -ge "$PGTIMEOUT" ]; then
        echo "."
        echo "Unable to reach PostgreSQL server" 1>&2
        exit 1
    fi
    echo -n "."
done

echo " [OK]"
```

vrenaville: Currently I have just revert, you fix, I will open a PR to use pg_isready as you describe

PR #267, opened 2024-01-04
fix bin/ utilities

gurneyalex:
local-src is no longer used -> things are now in /odoo/odoo/addons


simahawk (approved):
LG

gurneyalex:
I'm fixing the CI :-/

example/Dockerfile
```diff
@@ -8,9 +8,8 @@ MAINTAINER Camptocamp
 # 2. or adding each directory, this solution will reduce the build and download
 # time of the image on the server (layers are reused)
 RUN mkdir -p /odoo/src/odoo
-COPY ./src /odoo/src/odoo
-COPY ./external-src /odoo/external-src
-COPY ./local-src /odoo/local-src
+COPY ./odoo/src/odoo /odoo/src/odoo
```
yvaucher, line 11: Copying only ./odoo/src/odoo is ok, but you also need to copy ./odoo/src/setup.py to be able to pip install it.

gurneyalex: you are right, I thought I had misinterpreted https://github.com/camptocamp/docker-odoo-project/blob/master/test.sh#L66

yvaucher: Wait I missed your change on https://github.com/camptocamp/docker-odoo-project/pull/267/files#diff-3722d9ba8feb2d3feac8ce71a209a638d4b404e1c53f937188761181594023e2R67-R68

Which confused me.

So far I have not seen where the issue comes from. The path change seems to be the ideal culprit but I didn't find enough evidences to blame it yet. :judge:

gurneyalex:
I just rebased + squash

PR #282, opened 2024-08-29
feat: add chromium to support odoo tests running in a browser

nilshamerlinck:
Since [12.0](https://github.com/odoo/odoo/commit/de84ad9c1f4853cb8a748ee1d0c880081dae9fe2), odoo uses Chrome (Headless) instead of PhantomJS to run tests that needs a browser (e.g. Tour based tests for the POS).

On the OCA CI side, [Google Chrome is used](https://github.com/OCA/oca-ci/blob/f641d2587049ecd38637390f14db3bca3197a357/Dockerfile#L52).

But Odoo [also supports Chromium](https://github.com/odoo/odoo/blob/ba1ba88874e2ae9716e59ec199eb739d58c3f771/odoo/tests/common.py#L635), which is packaged in Debian.

So this PR adds `chromium` and the pip lib `websocket-client` that odoo uses to discuss with `chromium` through the Remote Debugging Protocol.

fyi:
```
$ apt show chromium | grep Installed-Size
Installed-Size: 235 MB
```

### DIFF
--- a/12.0/Dockerfile
+++ b/12.0/Dockerfile
@@ -27,7 +27,8 @@ ENV BUILD_PACKAGE \
     libjpeg-dev \
     libpng-dev \
     zlib1g-dev \
-    git
+    git \
+    chromium
 
 # Install some deps, lessc and less-plugin-clean-css, and wkhtmltopdf
 RUN set -x; \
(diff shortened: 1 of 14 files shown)


ivantodorovich:
I'd like to be able to run js tests, and for that we need this.

However this is not a production requirement, and it's a heavy one. I'm not sure it's worth shipping it in the production image.

nilshamerlinck:
Hi @ivantodorovich 

You're right, production image should not contain unnecessary deps.

I opened https://github.com/camptocamp/odoo-template/pull/752 as an alternative approach.

PR #299, opened 2024-10-15
[4.5.X] Refactor DB cache management in runtests and runmigration

yvaucher:
Backport of https://github.com/camptocamp/docker-odoo-project/pull/252 for the 4.5.X branch

Gives most variables a default value making them changable with var=${var:=default}

Extract db restore and dump in a separate db.sh file.

Change (default) value for cachedb directory to make it easier to set a volume on it.

Add || true in runmigration when doing ls -v to ignore the error status and set an empty CACHED_DUMP variable.

Improve slightly the messages.

This should fix issue of renaming in bin/runmigration of the dump file name which stayed the same as the restored one. (and in worst case to an empty filename on github actions)


yvaucher:
Rebasing on top of #300

PR #322, opened 2025-02-26
fix: switch back to pip install module

vrenaville:


16.0/base_requirements.txt
```diff
@@ -1,12 +1,87 @@
...
+vobject==0.9.6.1
+Werkzeug==2.1.1 ; python_version > '3.9'  # (Jammy)
+xlrd==1.2.0; python_version >= '3.8'
+XlsxWriter==1.1.2
+xlwt==1.3.0
+zeep==4.0.0
+
+setuptools==73.0.1
+
 # Not part of official requirements, but used by some addons
```
gurneyalex, line 49: I'm wondering why are those not in extra_requirements.txt?

16.0/base_requirements.txt
```diff
@@ -1,12 +1,87 @@
...
 # colorama==0.3.9
 gdata==2.0.18
-lxml-html-clean==0.4.1
-whool==1.2
-setuptools==75.6.0
```
gurneyalex, line 6: any reason to freeze setuptools to 73.x ?

16.0/base_requirements.txt
```diff
@@ -1,12 +1,87 @@
...
+mccabe==0.7.0
+more-itertools==10.5.0
+pbr==6.1.0
+pexpect==4.9.0
+ptyprocess==0.7.0
+py==1.11.0
+pycodestyle==2.12.1
+pyflakes==3.2.0
+unicodecsv==0.14.1
+wrapt==1.16.0
```
gurneyalex, line 87: maybe we could have a testing_requirements.txt to make things clearer, and maybe some day not include it in the production image?

gurneyalex (approved):

16.0/base_requirements.txt
vrenaville: Well, I'm a bit scary about setuptools, they are used to make breaking changes, that why I freeze it

PR #342, opened 2025-05-20
feat: add 14.0 version

vrenaville:


.github/workflows/build.yaml
```diff
@@ -29,7 +29,8 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        odoo_serie: ["15.0","16.0","17.0","18.0"]
+        #odoo_serie: ["14.0","15.0","16.0","17.0","18.0"]
+        odoo_serie: ["14.0"]
```
gurneyalex, line 33: don't we want to build for 15.0 -> 18.0 too?

PR #344, opened 2025-05-27
[FIX] add missing requirement for guessing mimetype

leemannd:
It is widely used in odoo to work around the mimetypes.
It was installed for `odoo-dj` into `odoo-template`.

I want to have it removed from odoo-template and added into the base images.

### DIFF
--- a/18.0/extra_requirements.txt
+++ b/18.0/extra_requirements.txt
@@ -38,6 +38,7 @@ pycparser==2.22
 pycryptodome==3.20.0
 PyNaCl==1.5.0
 pytesseract==0.3.13
+python3-magic
 regex==2024.7.24
 s3transfer==0.10.2
 tzlocal==5.2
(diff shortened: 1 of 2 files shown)


18.0/extra_requirements.txt
```diff
@@ -38,6 +38,7 @@ pycparser==2.22
 pycryptodome==3.20.0
 PyNaCl==1.5.0
 pytesseract==0.3.13
+python-magic
```
leemannd, line 41: Before merge I need to check also in older odoo versions

ivantodorovich (approved):
Yeap, Odoo's also installing it in their official image: https://github.com/odoo/docker/blob/5d67b9585c0e29c9dccc0b3f484eedc4e5c46714/18.0/Dockerfile#L25

leemannd:
Hello @ivantodorovich I think this should be started from scatch the files `extra_requirements` are not existing anymore.

Closing the PR

PR #360, opened 2025-09-18
imp: run-parts exit on error

ivantodorovich:
In case of error in a one of the parts/scripts, exit immediately.

Without this, the other parts will be executed, but it will still return an error code and so the docker-entrypoint.sh will immediately exit, too.

This way, though, it's easier to stop where the error is.

Motivated by:
-  https://github.com/camptocamp/docker-odoo-core/pull/302

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -8,6 +8,7 @@
 .. **Features and Improvements**
 
 * Replace environment variable OPENERP_SERVER with ODOO_RC
+* Entrypoint: exit on error when executing run-parts
 
 .. **Bugfixes**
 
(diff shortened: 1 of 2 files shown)


ivantodorovich:
> @ivantodorovich Thanks for you work, could you please rebase it should fix the pipeline

Sure thing, it's done!

vrenaville:
@ivantodorovich i think the issue came from secret env, it s not available on your fork I assume, it is possible to do a branch on the camptocamp repo instead ? then the env variable should be accessible

PS sorry for the long delay, I forgot to follow this topic

PR #429, opened 2026-04-08
ci: add Dockerfile linting with Hadolint

ivantodorovich:
Add Hadolint to CI and pre-commit to catch Dockerfile issues automatically.

Fixes for existing Dockerfile issues found by Hadolint:
- Replace deprecated MAINTAINER with LABEL maintainer= (DL4000)
- Add trailing slash to multi-arg COPY destinations (DL3021)
- Consolidate consecutive RUN instructions in example (DL3059)
- Use WORKDIR instead of cd in example (DL3003)

### DIFF
--- a/.github/workflows/hadolint.yaml
+++ b/.github/workflows/hadolint.yaml
@@ -1,0 +1,21 @@
+name: hadolint
+
+on:
+  push:
+    branches:
+      - "release-4.5.X"
+      - "dev-4.5.X"
+  pull_request:
+    branches:
+      - "release-4.5.X"
+
+jobs:
+
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Lint Dockerfiles
+        uses: hadolint/hadolint-action@v3.3.0
+        with:
+          recursive: true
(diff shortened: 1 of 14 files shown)


.github/workflows/hadolint.yaml
```diff
@@ -0,0 +1,21 @@
...
+      - "release-4.5.X"
+
+jobs:
+
+  hadolint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - name: Lint Dockerfiles
+        uses: hadolint/hadolint-action@v3.3.0
```
florentx, line 19: ### 1st question
Should we start pinning actions to commit SHA instead of tag, as recommended?

https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

### 2nd question

What is benefit of having it in a separate job, if it is already part of pre-commit hooks?

florentx (approved):
LG

.github/workflows/hadolint.yaml
ivantodorovich: > Should we start pinning actions to commit SHA instead of tag, as recommended?

Personally I prefer the version tags. But I don't have a strong opinion.
We're already doing this everywhere so if we decide to change, it should be globally

> What is benefit of having it in a separate job, if it is already part of pre-commit hooks?

pre-commit hooks are introduced by this PR as a way to easily run hadolint locally, but we're not running them in the CI -- instead we run the official hadolint gh action.

PR #431, opened 2026-04-14
fix: force dbfilter to False in the config file

gurneyalex:
This allows avoiding connection to the 'postgres' database to list the databases matching the filter. In case we need multiple databases, it is possible to have db_name set to a comma separated list. This works for Odoo >= 10.0


ivantodorovich (approved):

florentx:
- branch `release-4.5.X` is merged into `main`
- #432 is merged

PR #432, opened 2026-04-14
fix: force dbfilter to False in the config file

gurneyalex:
This allows avoiding connection to the 'postgres' database to list the databases matching the filter. In case we need multiple databases, it is possible to have db_name set to a comma separated list. This works for Odoo >= 10.0


ivantodorovich (approved):
❓ do we want to make updates on `master` branch? given it's deprecated and we're aiming to remove it

ivantodorovich:
In `19.0`, I'm getting the following warning:

```
WARNING ? odoo.tools.config: option dbfilter reads 'False' in the config file at /etc/odoo.cfg but isn't a boolean option, skip 
```

Shouldn't we, instead, simply remove the line?

gurneyalex:
> In `19.0`, I'm getting the following warning:
> 
> ```
> WARNING ? odoo.tools.config: option dbfilter reads 'False' in the config file at /etc/odoo.cfg but isn't a boolean option, skip 
> ```
> 
> Shouldn't we, instead, simply remove the line?

+1 -> https://github.com/camptocamp/docker-odoo-project/pull/458

PR #436, opened 2026-04-22
feat: split test requirements into `test_requirements.txt`

florentx:
This is the way images are built for 5.x branch.

### DIFF
--- a/14.0/base_requirements.txt
+++ b/14.0/base_requirements.txt
@@ -61,17 +61,6 @@ simplejson==3.19.3
 marabunta==0.12.0
 -e git+https://github.com/camptocamp/anthem@master#egg=anthem
 
-# test / lint
-# those libs and their dependencies are unpinned
-# to always test with the last version of it
-flake8
-pytest==7.4.4
-pluggy
-coverage
-pytest-odoo>=0.4.7
-pytest-cov>=2.10.0
-watchdog
-
 
 # Library dependency
 argh==0.31.3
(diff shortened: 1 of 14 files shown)


ivantodorovich (commented):
How are they installed then?

florentx:
> How are they installed then?

I'm looking closer, and it seems that it was removed from the 5.x branch by mistake.
First the Dockerfile for Odoo 19 was ported without installing test requirements:
 - #376 [link](https://github.com/camptocamp/docker-odoo-project/pull/376/changes#diff-683b729c04d7b2f020bfcd1b108c273f80e48ac3ff49fc2dac3257b89b4a2158)
 - then I propagated the mistake to other versions with #399 
 
So I will probably go for **proposal A**:
 - change current PR to keep installing `test_requirements` for 4.5.x

**Proposal B** would be to keep base Odoo image without the `test_requirements` installed, and only install these requirements on Dev environment?  (`DEV_MODE` set).
It would need a change in the `Dockerfile` of `odoo-template`, in this case.

ivantodorovich:
I like the idea to use DEV_MODE to have a slightly different image for dev and CI.

One thing we always missed in our images it to have chromium installed to be able to run javascript tests.
We never did that because it's not a good idea to include chromium in the production deployment.

But it's something that can be done conditionally with this DEV_MODE flag, I guess.

florentx:
Flag `DEV_MODE` is in `odoo-template`.
This can be a feature to develop later.

I propose to merge this PR, so we can progress on:
- #438

ivantodorovich (approved):

PR #439, opened 2026-04-28
feat: do not pre-install Postgres client 17

florentx:
Archive format generated with version 17 is not compatible with PG 14.

In brief:
- PostgreSQL client 12 to 15 are compatible together (archive format v1.14)
- PostgreSQL client 16 produces archive v1.15 (and read all previous)
- PostgreSQL client 17 and 18 produce archive v1.16 (and read older)

This change partially reverts:
- #418 

Reference:
- https://github.com/postgres/postgres/blob/REL_18_STABLE/src/bin/pg_dump/pg_backup_archiver.h#L67

Trick, to print the headers of a PG dump:
```sh
$ pg_restore -ln- path/to/database-dump.pg
;
; Archive created at 2026-02-03 18:41:36 CET
;     dbname: broken_shadow_592
;     TOC Entries: 17387
;     Compression: gzip
;     Dump Version: 1.14-0
;     Format: CUSTOM
;     Integer: 4 bytes
;     Offset: 8 bytes
;     Dumped from database version: 14.19
;     Dumped by pg_dump version: 14.17
;
;
; Selected TOC Entries:
;
```

### DIFF
--- a/12.0/Dockerfile
+++ b/12.0/Dockerfile
@@ -35,7 +35,7 @@ ENV BUILD_PACKAGE="\
 RUN set -x; \
     /install/package_odoo.sh \
     && /install/setup-pip.sh \
-    && /install/postgres.sh \
+    && /install/postgres.sh 15 \
     && /install/kwkhtml_client.sh \
     && /install/kwkhtml_client_force_python3.sh \
     && /install/dev_package.sh \
--- a/13.0/Dockerfile
+++ b/13.0/Dockerfile
@@ -35,7 +35,7 @@ ENV BUILD_PACKAGE="\
 RUN set -x; \
     /install/package_odoo.sh \
     && /install/setup-pip.sh \
-    && /install/postgres.sh \
+    && /install/postgres.sh 15 \
     && /install/kwkhtml_client.sh \
     && /install/kwkhtml_client_force_python3.sh \
     && /install/dev_package.sh \
--- a/README.md
+++ b/README.md
@@ -76,12 +76,17 @@ You can also see the Dockerfile that generated this image here: [common/Dockerfi
 The images should be built with `make`:
 
 Normal flavors:
-
 ```
 $ make VERSION=19.0
 ```
 
-Batteries-included flavors:
+Note: version of PostgreSQL client can be changed.  By default, it is version 15, which is fully
+compatible with PostgreSQL 12 to 15. Example:
+```
+$ make VERSION=19.0 PG_VERSION=17
+```
+
+Batteries-included flavor:
 ```
 $ make VERSION=19.0 BATTERIES=True
 ```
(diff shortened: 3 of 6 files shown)


ivantodorovich:
Odoo.sh is using:
```
psql --version
psql (PostgreSQL) 16.13
```

If we use psql client < 16, it means we won't be able to restore an Odoo.sh archive (archive v1.15)?


> Archive format generated with version 17 is not compatible with PG 14.

Isn't it only true for another postgres client, but still pretty much compatible with PG server v14 (as long as the db itself s not using incompatible server features, which should not be the case with Odoo)?

ivantodorovich (approved):
Ok I see now the other PR, and that we were already doing this directly in odoo-template. So it should be ok

We should check if the incompatibility with the c2c-cloud-platform is still valid, though.

florentx:
> If we use psql client < 16, it means we won't be able to restore an Odoo.sh archive (archive v1.15)?
> 

There is possibility to generate an image for PG version 16 (Odoo.sh).  I've added in the README.
Maybe we could extend CI to generate these images periodically too? It should be easy.

> > Archive format generated with version 17 is not compatible with PG 14.
> 
> Isn't it only true for another postgres client, but still pretty much compatible with PG server v14 (as long as the db itself s not using incompatible server features, which should not be the case with Odoo)?

What is relevant is the version of `pg_dump`/`pg_restore` that you are using.
- :white_check_mark:  if server is 14 and tools `pg_dump` /`pg_restore` are version 17 --> it work fine
- :white_check_mark:  if server is 14 and `pg_dump` is 14 and `pg_restore` is 17 --> it works fine
- :x:  if server is 14 and `pg_dump` is 17 and `pg_restore` is 14 --> KO.  `pg_restore` command will complain and refuse to restore.  This is certainly the issue which was reported as "incompatibility with the c2c-cloud-platform".

Also good to know:
- :white_check_mark:  if server is 14 and `pg_dump` is 15 and `pg_restore` is 12 --> it works fine since the dump as format v1.14, which is accepted by `pg_restore` 12 to 15
- :x:  if server is 14 and `pg_dump` is version 12 --> KO even if archive format is v1.14 for both
```
pg_dump: error: server version: 14.13 (Debian 14.13-1.pgdg120+1); pg_dump version: 12.22 (Ubuntu 12.22-2.pgdg20.04+1)
pg_dump: error: aborting because of server version mismatch
```

ivantodorovich:
Yes that's exactly my point. Perhaps the real solution is not to downgrade the postgres client here, but to bump the postgres client used by our platform tooling -- that would get us:
- no multiple flavors needed
- compatible with odoo.sh / other external dumps


(not blocking this, though -- it's already approved. just mention this for the future)


----

related to: https://camptocamp.atlassian.net/browse/RUN-2120

PR #445, opened 2026-04-29
[FIX] Do not compare value when checking if 'database.is_neutralized'

twalter-c2c:
`odoo neutralize` set this parameter to `true` (see below).
Instead of checking if the value is `True`, as we did,
we can only check if the parameter exists to not care about
its exact value.
```
-- neutralization flag for the database
INSERT INTO ir_config_parameter (key, value)
VALUES ('database.is_neutralized', true)
    ON CONFLICT (key) DO
       UPDATE SET value = true;
```

### DIFF
--- a/start-entrypoint.d/002_neutralize
+++ b/start-entrypoint.d/002_neutralize
@@ -11,7 +11,7 @@ if [[ "$RUNNING_ENV" != "prod" && "$ODOO_NEUTRALIZE" == "True" ]]; then
         echo "Database does not exist, ignoring script"
         exit 0
     fi
-    if [ "$( psql -tAc "SELECT value FROM ir_config_parameter WHERE key='database.is_neutralized'" )" = 'True' ]; then
+    if [ "$( psql -tAc "SELECT 1 FROM ir_config_parameter WHERE key='database.is_neutralized'" )" = '1' ]; then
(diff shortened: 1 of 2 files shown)


twalter-c2c:
Closing in favor of https://github.com/camptocamp/docker-odoo-project/pull/446 since I forgot to not use fork when creating a PR.

start-entrypoint.d/002_neutralize
```diff
@@ -11,7 +11,7 @@ if [[ "$RUNNING_ENV" != "prod" && "$ODOO_NEUTRALIZE" == "True" ]]; then
         echo "Database does not exist, ignoring script"
         exit 0
     fi
-    if [ "$( psql -tAc "SELECT value FROM ir_config_parameter WHERE key='database.is_neutralized'" )" = 'True' ]; then
+    if [ "$( psql -tAc "SELECT 1 FROM ir_config_parameter WHERE key='database.is_neutralized'" )" = '1' ]; then
```
florentx, line 14: Maybe check if parameter is empty?  this is what is done by Odoo's `<ir.config_parameter>.get_param`.


```suggestion
    if [ "$( psql -tAc "SELECT 1 FROM ir_config_parameter WHERE key='database.is_neutralized' AND value != '';" )" = '1' ]; then
```

PR #447, opened 2026-05-15
Make orjson available

gurneyalex:
In Odoo >= 17.0 this is used in the bus addon for speed up.

See https://github.com/odoo/odoo/commit/62accdaea193a59c19d35ae5b79d3be1f77dfdce

### DIFF
--- a/17.0/base_requirements.txt
+++ b/17.0/base_requirements.txt
@@ -56,6 +56,7 @@ html5lib==1.1
 odfpy==1.4.1
 pyinotify==0.9.6
 simplejson==3.19.3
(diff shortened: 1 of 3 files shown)


florentx (approved):
Looks good.

I believe that `simplejson` requirement is a vestige from Odoo 9.0 or something like this.

IMHO, it can be dropped.

leemannd (approved):

divad1196:
@gurneyalex Good catch, it's easy to miss. 
Maybe we should make a PR to odoo to add it has an extra dependency of setup.py ?
https://github.com/odoo/odoo/blob/19.0/setup.py#L71

PR #448, opened 2026-05-18
fix: remove obsolete simplejson dependency

florentx:
This was removed upstream in Odoo 9: 
- odoo/odoo#6940

And keeping 'simplejson' around can have side effects:
- odoo/odoo#264303

(see also #447)


leemannd (approved):

twalter-c2c (approved):

divad1196 (approved):

divad1196:
@florentx simplejson is not useful for python 2.7 and above. Surprised it stayed until today.
Good catch.

PR #453, opened 2026-05-27
fix(apt_purge): pass packages as separate args to apt-get remove

vvrossem:
`$@` inside a double-quoted default-assignment `${var=$@}` is joined into a single string, causing `apt-get remove` to receive all package names as one argument instead of individual ones.
Replace the pattern with a plain argument-count check (`$# -gt 0`) and pass `"$@"` directly so each package remains a separate word.

### DIFF
--- a/install/apt_purge.sh
+++ b/install/apt_purge.sh
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -eo pipefail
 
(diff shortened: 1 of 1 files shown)


florentx (approved):
The fix looks OK.

It is weird that I did not have the issue.

And base Docker image builds without error, and without the fix.

PR #461, opened 2026-06-11
feat: provide an immutable unaccent pg function

gurneyalex:
Tests I did: 

* local test on alloboissons rebuilt with this image: get a production database, run odoo twice, check that the second time the fix index action does nothing
* local test on alloboissons project: start with no database, check that the `unaccent` function is created and that the installation of the modules creates the expected database indexes.

### DIFF
--- a/before-migrate-entrypoint.d/100_setup_unaccent
+++ b/before-migrate-entrypoint.d/100_setup_unaccent
@@ -1,0 +1,36 @@
+#!/bin/bash
+if [ "$( psql -tAc "SELECT 1 FROM pg_database WHERE datname='$DB_NAME'" )" != '1' ]
+then
+    # we don't want odoo to automatically create the database,
+    # because this would not initialize unaccent as we want
+    createdb -O ${DB_USER} ${DB_NAME}
+fi
+
+psql ${db} << EOF
+
+-- cleanup
+  DO \$do\$
+  BEGIN
+    IF EXISTS (
+      SELECT 1
+      FROM pg_extension e
+      JOIN pg_namespace n ON n.oid = e.extnamespace
+      WHERE e.extname = 'unaccent'
+        AND n.nspname = 'public'
+    ) THEN
+      DROP EXTENSION unaccent;
+    END IF;
+  END
+  \$do\$;
+
+  CREATE SCHEMA IF NOT EXISTS unaccent_schema; 
+  CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA unaccent_schema; 
+  CREATE OR REPLACE FUNCTION public.unaccent(text) 
+  RETURNS text 
+  LANGUAGE sql 
+  IMMUTABLE PARALLEL SAFE STRICT 
+  AS \$function\$ 
+     SELECT unaccent_schema.unaccent('unaccent_schema.unaccent', \$1) 
+  \$function\$ ;
+EOF
+
--- a/before-migrate-entrypoint.d/100_setup_unaccent
+++ b/before-migrate-entrypoint.d/100_setup_unaccent
@@ -1,36 +1,37 @@
 #!/bin/bash
+set -x
+
+# Odoo uses "unaccent" function for searches.  It is provided by PG extension.
+# However, this function needs to be "immutable" to be able to create indexes.
+#
+# When new database is created, Odoo tries to set the function immutable.
+# However it might not work in case the PostgreSQL server is hosted on a cloud platform.
+#
+# This is the reason for this workaround.
+#
+# Reference:
+# - https://github.com/odoo/odoo/blob/19.0/odoo/service/db.py#L155-L163
+# - https://github.com/odoo/odoo/pull/95943
+
 if [ "$( psql -tAc "SELECT 1 FROM pg_database WHERE datname='$DB_NAME'" )" != '1' ]
 then
-    # we don't want odoo to automatically create the database,
-    # because this would not initialize unaccent as we want
-    createdb -O ${DB_USER} ${DB_NAME}
+    echo "Database does not exist, ignoring script"
+    exit 0
 fi
 
-psql ${db} << EOF
+# If 'unaccent' is already IMMUTABLE, do nothing
+if [ "$( psql -tAc "SELECT 1 FROM pg_proc WHERE proname = 'unaccent' AND provolatile = 'i' AND pronamespace = current_schema::regnamespace;" )" = "1" ]
+then
+    echo "Function 'unaccent' is already immutable. Skip"
+else
+    psql << EOF
+DROP EXTENSION IF EXISTS unaccent;
+CREATE SCHEMA IF NOT EXISTS unaccent_schema;
+CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA unaccent_schema;
 
--- cleanup
-  DO \$do\$
-  BEGIN
-    IF EXISTS (
-      SELECT 1
-      FROM pg_extension e
-      JOIN pg_namespace n ON n.oid = e.extnamespace
-      WHERE e.extname = 'unaccent'
-        AND n.nspname = 'public'
-    ) THEN
-      DROP EXTENSION unaccent;
-    END IF;
-  END
-  \$do\$;
-
-  CREATE SCHEMA IF NOT EXISTS unaccent_schema; 
-  CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA unaccent_schema; 
-  CREATE OR REPLACE FUNCTION public.unaccent(text) 
-  RETURNS text 
-  LANGUAGE sql 
-  IMMUTABLE PARALLEL SAFE STRICT 
-  AS \$function\$ 
-     SELECT unaccent_schema.unaccent('unaccent_schema.unaccent', \$1) 
-  \$function\$ ;
+CREATE OR REPLACE FUNCTION unaccent(value text) RETURNS text
+LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT
+RETURN (SELECT unaccent_schema.unaccent('unaccent_schema.unaccent', value));
 EOF
 
+fi
--- a/start-entrypoint.d/100_fix_indexes
+++ b/start-entrypoint.d/100_fix_indexes
@@ -1,0 +1,36 @@
+#! /bin/bash
+# This script can be removed when we have finished updating databases with
+# the configuration of the unaccentextension (odoo >= 16.0)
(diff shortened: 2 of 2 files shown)


before-migrate-entrypoint.d/100_setup_unaccent
```diff
@@ -0,0 +1,36 @@
...
+        AND n.nspname = 'public'
+    ) THEN
+      DROP EXTENSION unaccent;
+    END IF;
+  END
+  \$do\$;
+
+  CREATE SCHEMA IF NOT EXISTS unaccent_schema; 
+  CREATE EXTENSION IF NOT EXISTS unaccent WITH SCHEMA unaccent_schema; 
+  CREATE OR REPLACE FUNCTION public.unaccent(text)
```
ivantodorovich: If I'm reading this right, this will re-create (due to `OR REPLACE`) the function on each container restart
Not sure if that's a problem, maybe not

gurneyalex: It should not be a problem: the syntax will not break existing indexes (dropping and recreating could break them), and the function update in itself is fast. 

we could query `pg_routines` to look for it, but I'm not sure this  would gain anything.

before-migrate-entrypoint.d/100_setup_unaccent
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+if [ "$( psql -tAc "SELECT 1 FROM pg_database WHERE datname='$DB_NAME'" )" != '1' ]
+then
+    # we don't want odoo to automatically create the database,
+    # because this would not initialize unaccent as we want
+    createdb -O ${DB_USER} ${DB_NAME}
```
gurneyalex: Feedback on this is welcome

before-migrate-entrypoint.d/100_setup_unaccent
ivantodorovich: As long as it doesn't break the indexes it should be fine like this 👌🏻

before-migrate-entrypoint.d/100_setup_unaccent
ivantodorovich: IMO we don't need it.

In our platform the db will always be created beforehand, so it's not an issue.

I guess this is only for our local.. but we don't need this schema workaround there. In local you have power rights on psql so the [Odoo-way already there works just fine](https://github.com/odoo/odoo/blob/2e2a4180106d28a477798d8011de6f07d1a58615/odoo/service/db.py#L155-L163) -- following restarts on local will have this entrypoint create the schema and re-create the function, but it shouldn't be any issue.

start-entrypoint.d/100_fix_indexes
```diff
@@ -0,0 +1,36 @@
...
+    print(f"Fixing index {indexname} on table {tablename}")
+    cr.execute(f"DROP INDEX {indexname}")
+if tables:
+    # recreate the indexes with unaccent
+    models = set()
+    for model in self.env.registry.models.values():
+        if model._table in tables:
+            models.add(model._name)
+    self.env.registry.check_indexes(self.env.cr, models)
+self.env.cr.commit()
```
ivantodorovich: This script is too version specific IMO to keep it under a >=16.0 condition.
It can easily crash in X years time.

I'm not sure the docker image is the right place for it, but I can't think of anything better so... let's keep it

But maybe we could cap it to 19.0, cause won't need a migration path for newly created future 20.0+ projects.

before-migrate-entrypoint.d/100_setup_unaccent
gurneyalex: good point. I'll check if things work well with an early exit path here.

florentx: If we keep this part, we could add comment, that it skips Odoo configuration: `db_template`

And the logic is slightly different than the Odoo way (no template, no LC_COLLATE).
https://github.com/odoo/odoo/blob/19.0/odoo/service/db.py#L142-L148

before-migrate-entrypoint.d/100_setup_unaccent
```diff
@@ -0,0 +1,36 @@
...
+fi
+
+psql ${db} << EOF
```
florentx: As a drop-in replacement, a bit simpler, I've tested this:

```sh
if [ "$( psql -tAc "SELECT 1 FROM pg_proc WHERE proname = 'unaccent' AND provolatile = 'i' AND pronamespace = current_schema::regnamespace;" )" = "1" ]
then
    echo "Function 'unaccent' is already immutable. Skip"
else
    psql << EOF
DROP EXTENSION IF EXISTS unaccent;
CREATE SCHEMA IF NOT EXISTS unaccent_schema;
CREATE EXTENSION unaccent WITH SCHEMA unaccent_schema;

CREATE OR REPLACE FUNCTION unaccent(value text) RETURNS text
LANGUAGE sql IMMUTABLE PARALLEL SAFE STRICT
RETURN (SELECT unaccent_schema.unaccent('unaccent_schema.unaccent', value));
EOF
```

Syntax is supported for PostgreSQL 14+

start-entrypoint.d/100_fix_indexes
florentx: Variant of the script, which checks if the index name is similar to index created by `tools.sql.make_index_name` so it should not drop other custom indexes which are project-specific.

```py
if [ "${ODOO_VERSION%%.*}" -lt 16 ]; then
    echo "Index fixing for unaccent: requires Odoo 16.0+"
    exit 0
fi

WHERE_CLAUSE="WHERE indexdef LIKE '%gin_trgm_ops%' AND indexdef NOT LIKE '%unaccent%' AND schemaname = current_schema"

if [ 0 = $(psql -tAc "SELECT COUNT(*) FROM pg_indexes $WHERE_CLAUSE;") ]
then
    echo "No indexes to fix"
    exit 0
fi

odoo shell --no-http --log-handler=:ERROR --log-handler=odoo.schema:INFO << EOF
import re

query = """SELECT indexname, tablename FROM pg_indexes $WHERE_CLAUSE;"""
env.cr.execute(query)

tables = set()
for indexname, tablename in env.cr.fetchall():
    # Odoo 17+ uses '__' as separator between table and column
    # Note that index names longer than 63 chars are mangled, hence they are not detected
    parsed = re.match(f"{tablename}__?(.+)_index", indexname)

    # Sanity check, to avoid dropping non-standard indexes
    colname = parsed and parsed.group(1)
    if not colname:
        print(f"Left unchanged: {indexdef}")
        continue
    tables.add(tablename)
    print(f"Fixing index {indexname} on table {tablename} ({colname})")
    env.cr.execute(f"DROP INDEX {indexname};")

if tables:
    # recreate the indexes with unaccent
    models = set()
    for model in env.registry.models.values():
        if model._table in tables:
            models.add(model._name)
    env.registry.check_indexes(env.cr, models)
env.cr.commit()
EOF
```

florentx (commented):
I would suggest to add a note in the `before-migrate` script, which explain why it is needed, and why Odoo's way is not working on cloud platform.
https://github.com/odoo/odoo/pull/95943/changes#diff-f23ced3ea41a0a5593df3d5a67533aa3b433080e2a7a33f81d4df36502382fe0


Also I propose to change, not to hard-code namespace `public`, since some projects have peculiar constraints.

start-entrypoint.d/100_fix_indexes
florentx: FTR, this is where the index names were changed (Odoo 17):
- odoo/odoo#100736 this [part](https://github.com/odoo/odoo/pull/100736/changes#diff-5f043f5617daedcdc16b95f0f4bf3f2865417f03e18b4b33f9afc86cb128599a)

before-migrate-entrypoint.d/100_setup_unaccent
ivantodorovich: I like that we check if the extension is already immutable instead 💯

start-entrypoint.d/100_fix_indexes
gurneyalex: I would be tempted to assume that manually created indexes should be mirrored in the code. At least this is my practice when I do manual tuning.

before-migrate-entrypoint.d/100_setup_unaccent
gurneyalex: Agreed.

before-migrate-entrypoint.d/100_setup_unaccent
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+if [ "$( psql -tAc "SELECT 1 FROM pg_database WHERE datname='$DB_NAME'" )" != '1' ]
+then
+    # we don't want odoo to automatically create the database,
+    # because this would not initialize unaccent as we want
```
divad1196: I think that we are missing the "why" the automatic creation not "as we want".
I found [our doc](https://camptocamp.atlassian.net/wiki/spaces/DEV/pages/1290272781/Unaccent+support+in+Odoo+and+the+cloud+platform) with the same script and had to track the information 
1. We want unaccent in a different schema
2. Because Odoo user cannot alter it to make it immuable

But it's not explained why we are in this situation in the first place, not in the script nor in the doc.

## Missing information
From my researches, I found the following information that should IMO be added to the documentation / this script's comments

#### Why Odoo need immutable `unaccent`
Odoo needs the `unaccent` function for the indexes, but postgres only allow to use immutable function for that. Therefore Odoo needs `unaccent` to be immutable.

#### Why can't odoo edit `unaccent`
I assume that it's because we have shared cluster, so Odoo user isn't a superuser.
Extension is being installed by a different user.

#### Why don't we set the immutable flag
Because odoo will run `cr.execute("ALTER FUNCTION unaccent(text) IMMUTABLE")` blindly,
But this only happens in `_create_empty_database`

=> I agree with @ivantodorovich: it's only running locally, but locally Odoo is superuser so no workaround needed. I couldn't find a definitive reason for this script. This should be documented if there is a reason for it.

## Documentation update

Is [this guide](https://camptocamp.atlassian.net/wiki/spaces/BS/pages/13965184/Use+UNACCENT+in+a+project) still valid? In particular the `Activating in our projects` section ?

start-entrypoint.d/100_fix_indexes
florentx: Yes it is good practice to have the indexes in code. But I don't think that `env.registry.check_indexes` will be enough to recreate the custom indexes.

florentx (approved):
Updated with suggestions

ivantodorovich (approved):

PR #467, opened 2026-07-27
BSRD-850: imp: always neutralize once when RUNNING_ENV is dev

ivantodorovich:
## Problem

`ODOO_NEUTRALIZE` skips databases that are already flagged as neutralized (`database.is_neutralized`).

A database restored locally almost always comes from another environment (integration, lab, ...) where it *was* already neutralized. So the check makes us skip the neutralization entirely, and the database keeps the values of the environment it came from: mail servers, crons, payment providers, ...

## Solution

On `RUNNING_ENV=dev`, guard the neutralization with a dedicated `database.is_neutralized_for_dev` system parameter instead of `database.is_neutralized`:

* a database restored from anywhere is neutralized once, locally;
* subsequent starts don't neutralize again, since it's not needed anymore.

The flag is only written on `dev`. On the other environments the flag set by `odoo neutralize` is enough, and writing ours there would end up inside the dumps, defeating the purpose.

Behavior on `prod` (never neutralized) and on the other non-prod environments (skipped when `database.is_neutralized` is set) is unchanged. `ODOO_NEUTRALIZE=False` is still the way to opt out.

Also replaced the explicit error propagation by `set -e`, so the flag can't be written if `odoo neutralize` fails — the neutralization is then retried on the next start.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,8 @@ steps on top of it.
 
 `ODOO_NEUTRALIZE` can be `True` or `False`. When set to `True` (default), the image will execute the official Odoo neutralization script if `RUNNING_ENV` is not set to `prod`. This deactivates outgoing emails, scheduled actions, and external payments to prevent accidental data leaks from production dumps.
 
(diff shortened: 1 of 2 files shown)


ivantodorovich:
Related to BSRD-850's _"Ensure neutralize is always executed in local dev"_

I believe this is not enough to guarantee it, but it's needed

gurneyalex (approved):

twalter-c2c (approved):

PR #472, opened 2026-07-30
fix: allow the command to be called with gosu odoo

gurneyalex:
Without this fix, the start-entrypoint.d/ scripts are not run when the command of the container is something similar to 'gosu odoo' (whereas the before-migrate-entrypoint.d/ scripts are run). This is the case if the Chart used is >= 5.3.8.

--

Related change:
- camptocamp/odoo-platform-charts#166

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -98,7 +98,8 @@ if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "m
     run-parts --exit-on-error --verbose "$BEFORE_MIGRATE_ENTRYPOINT_DIR"
   fi
 fi
-if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ]; then
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] ||
+    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ "odoo migrate" ]] ); then
 
   # Bypass migrate when `odoo shell` or `odoo --help` are used
   if [[ ! " ${ARGS[@]} " =~ " --help " ]] && [[ ! " ${ARGS[@]:0:1} " =~ " shell " ]]; then


bin/docker-entrypoint.sh
```diff
@@ -98,7 +98,8 @@ if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "m
     run-parts --exit-on-error --verbose "$BEFORE_MIGRATE_ENTRYPOINT_DIR"
   fi
 fi
-if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ]; then
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] ||
+    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ "odoo migrate" ]] ); then
```
florentx, line 102: It shoud be "odoo odoo" in this case:



```suggestion
    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ ^(odoo odoo) ]] ); then
```

bin/docker-entrypoint.sh
florentx, line 95: This needs to be changed as well, in order to capture both `gosu odoo odoo ...` and `gosu odoo migrate ...`.


```suggestion
    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ ^(odoo (odoo|migrate)) ]] ); then
```

bin/docker-entrypoint.sh
```diff
@@ -98,7 +98,8 @@ if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "m
...
 
   # Bypass migrate when `odoo shell` or `odoo --help` are used
   if [[ ! " ${ARGS[@]} " =~ " --help " ]] && [[ ! " ${ARGS[@]:0:1} " =~ " shell " ]]; then
```
florentx, line 105: Should we also take into account `gosu odoo` variant here?


```suggestion
  if [[ ! " ${ARGS[@]} " =~ " --help " ]] && [[ ! "${ARGS[@]}" =~ ^shell ]] && ! ( [ $BASE_CMD = "gosu" ] && [[ "${ARGS[@]}" = ^(odoo odoo shell) ]] ); then
```

florentx:
I've done a simpler refactoring:
- #473

florentx (changes requested):

gurneyalex:
superseeded by #473

PR #473, opened 2026-07-30
fix: better detect when command is run as `gosu odoo`

florentx:
Simpler fix for the issue with Chart 5.3.8 / 5.3.9

Replaces:
- #472

---

Without this fix, the start-entrypoint.d/ scripts are not run when the command of the container is something similar to 'gosu odoo' (whereas the before-migrate-entrypoint.d/ scripts are run). This is the case if the Chart used is >= 5.3.8.

--

Related change:
- camptocamp/odoo-platform-charts#166

### DIFF
--- a/bin/docker-entrypoint.sh
+++ b/bin/docker-entrypoint.sh
@@ -91,8 +91,7 @@ if [ $EUID -eq 0 ]; then
   fi
 fi
 
-if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "migrate" ] ||
-    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ "odoo migrate" ]] ); then
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "migrate" ]; then
   BEFORE_MIGRATE_ENTRYPOINT_DIR=$ODOO_BASE_PATH/before-migrate-entrypoint.d
   if [ -d "$BEFORE_MIGRATE_ENTRYPOINT_DIR" ]; then
     run-parts --exit-on-error --verbose "$BEFORE_MIGRATE_ENTRYPOINT_DIR"
--- a/common/env_odoo.sh
+++ b/common/env_odoo.sh
@@ -7,5 +7,11 @@ ODOO_BASE_PATH=""
 USER_ID=${LOCAL_USER_ID:-999}
 id -u odoo &> /dev/null || useradd --shell /bin/bash -u $USER_ID -o -c "" -m odoo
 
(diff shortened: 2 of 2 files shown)


bin/docker-entrypoint.sh
```diff
@@ -91,8 +91,7 @@ if [ $EUID -eq 0 ]; then
   fi
 fi
 
-if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "migrate" ] ||
-    ([ "$BASE_CMD" = "gosu" ] && [[ "${ARGS[@]}" =~ "odoo migrate" ]] ); then
+if [ "$BASE_CMD" = "odoo" ] || [ "$BASE_CMD" = "odoo.py" ] || [ "$BASE_CMD" = "migrate" ]; then
```
divad1196, line 94: Relying on the command as parameter is a bit unstable, even with `basename`.

Can we invert the logic to define when to NOT run parts instead?
```bash
if [ "$BASE_CMD" != "bash" ] && ...;
```

and/or can we add safe guards in `migrate` and `odoo` scripts ?
```bash
if [ "$BEFORE_MIGRATE_DONE" != "1" ]; then
    echo "migration didn't run, there is a bug in .." >&2
    exit 1
fi
```
version upgrade will be tested and such issue must not be silent during the tests

gurneyalex: There are many places when we don't want to run the scripts (all the postgresql commands, lots of bash commands...) and a few where we want to run, so the current logic seems better to me. 

I often use `docker compose run --rm odoo ls/cat/grep` for instance

gurneyalex (approved):

bin/docker-entrypoint.sh
divad1196: @gurneyalex I assumed this would be reason. 
Anonymizing should be the default, runing free command that by-pass anonymisation should explicitly bypass it.
This condition is a black box that increase the risk of bugs and makes it harder to reason about.

Since this one is going to be a long topic, I proposed adding a guard for this command to ensure the scripts ran.
What about this proposal?

florentx: I tried to look how to add a safeguard, but there's no obvious solution.

There are 2 interesting paths when `docker-entrypoint.sh` is executed

1. Command `migrate` or `gosu odoo migrate`:
 - 1. parts `before-migrate-entrypoint.d` executed
 - 2. `bin/migrate` executed

2. Command `odoo ...` or `gosu odoo odoo ...`:
 - 1. parts `before-migrate-entrypoint.d` executed
 - 2. `bin/migrate` executed
 - 3. parts `start-entrypoint.d` executed
 - 4. `odoo` started
 
 So the safeguard in `migrate` could only detect if "before" entrypoints are executed.
 And adding a safeguard to `odoo` would need to patch standard Odoo. Not a good idea.

PR #476, opened 2026-08-26
feat: tidy-up requirements

florentx:
Changes:
 - use same 'src_requirements.txt' for all Odoo versions
 - use same 'test_requirements.txt' for all Odoo versions
 - require recent 'openupgradelib' 3.13.7
 - require 'python-slugify' for fs_storage and remove obsolete 'slugify'
 - check requirements compatiblity in test


leemannd (approved):

divad1196 (commented):
For `src_requirements`, I thought it was a pending discussion.

In particular, I see that all version now use the exact same file, but previous version had different requirements (e.g. crypto libraries).

- Can they really have the same versions ?
- Are we going to have to template the file for conditional dependencies ?
  Because, IMO, this should be avoided.

florentx:
> For `src_requirements`, I thought it was a pending discussion.
> 
> In particular, I see that all version now use the exact same file, but previous version had different requirements (e.g. crypto libraries).
> 
>     * Can they really have the same versions ?
> 
>     * Are we going to have to template the file for conditional dependencies ?
>       Because, IMO, this should be avoided.

I think that you mix-up with `base_requirements.txt` which is still specific per version

PR #477, opened 2026-08-26
fix: caching dump for tests and migrations

florentx:
Port of changes from:
- camptocamp/odoo-ci-workflows#25

And fix condition for `echo "No cached migration sample dump found"` which was inverted.

### DIFF
--- a/bin/runmigration
+++ b/bin/runmigration
@@ -20,16 +20,14 @@
 #   if set to "false", will skip trying to reload cached dump named "odoo_sample_$VERSION.dmp"
 #
 # MIG_LOAD_VERSION_CEIL:
-#   Version number passed to search for an older dump.
+#   Version number passed to search for a dump.
 #
-#   It will load a dump lower than "odoo_sample_$MIG_LOAD_VERSION_CEIL.dmp"
-#   This is useful if you bumped odoo/VERSION as it won't match existing
-#   dumps.
+#   It will load a dump lower than or equal to "odoo_sample_$MIG_LOAD_VERSION_CEIL.dmp"
 #
-#   For instance you have made a dump 10.1.0, you are now on the version
-#   10.2.0, if you pass your current version it will search for a dump
-#   lower than 10.2.0 and restore the 10.1.0. Then play the remaining
-#   steps on top of it.
+#   For instance you have made a dump 10.1.0, you are now preparing version 10.2.0
+#   but odoo/VERSION will be bumped when the release is tagged. So passing 10.1.0
+#   it will return the dump you've made before and restore 10.1.0.
+#   Then play the remaining steps on top of it.
 set -e
 
 . db_cache_helpers.sh
@@ -42,19 +40,13 @@ wait_postgres.sh
 VERSION=$(cat /odoo/VERSION)
 
 if [ "$LOAD_DB_CACHE" != "false" ]; then
-
-  # If we want to run the migration steps on top of a previous dump
-  # useful when odoo/VERSION was edited
-  if [ -n "$MIG_LOAD_VERSION_CEIL" ]; then
-    echo "New version - Searching for previous version dump 🔭"
-    if [ -d "$CACHE_DIR" ]; then
-      # Filter dumps of higher releases
-      export MAX_DUMP="$CACHE_DIR/odoo_sample_${MIG_LOAD_VERSION_CEIL}.dmp"
-      CACHED_DUMP=$(ls -v $CACHE_DIR/odoo_sample_*.dmp || true | awk '$0 < ENVIRON["MAX_DUMP"]' | tail -n1)
-    fi
-    if [ -n "$CACHED_DUMP" ]; then
-      echo "No cached migration sample dump found"
-    fi
+  if [ -d "$CACHE_DIR" ] && [ -z "$CACHED_DUMP" ]; then
+    echo "Searching for previous version dump 🔭"
+    # Run the migration steps on top of a previous dump, filter out dumps of higher releases
+    CACHED_DUMP=$(ls -v $CACHE_DIR/odoo_sample_*.dmp || true | cap_version ${MIG_LOAD_VERSION_CEIL-} | tail -n1)
+  fi
+  if [ -z "$CACHED_DUMP" ]; then
+    echo "No cached migration sample dump found"
   fi
 else
   echo "DB cache load disabled."
@@ -62,7 +54,7 @@ fi
 
 if [ "$LOAD_DB_CACHE" != "false" ]; then
   db_restore "$CACHED_DUMP" "$DB_NAME"
-  if [ $? = 1 ]; then
+  if [ -f "$CACHED_DUMP" ]; then
     echo "Do migration on top of restored dump"
   else
     echo "Do migration from scratch 🐢 🐢 🐢"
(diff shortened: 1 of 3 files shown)


divad1196:
@florentx Thank you! I should have done it long ago, sorry :pray: 

I will review it when I have time since there are a few adaptations

bin/runmigration
```diff
@@ -42,27 +40,28 @@ wait_postgres.sh
 VERSION=$(cat /odoo/VERSION)
 
 if [ "$LOAD_DB_CACHE" != "false" ]; then
-
-  # If we want to run the migration steps on top of a previous dump
-  # useful when odoo/VERSION was edited
-  if [ -n "$MIG_LOAD_VERSION_CEIL" ]; then
-    echo "New version - Searching for previous version dump 🔭"
-    if [ -d "$CACHE_DIR" ]; then
+  if [ -d "$CACHE_DIR" -a -z "$CACHED_DUMP"]; then
```
divad1196: 2 things:

### Bug
missing space before `]`

### Syntax
`[ -d "$CACHE_DIR" ] && [ -z "$CACHED_DUMP" ]`

In the `APPLICATION USAGE` section,
> The XSI extensions specifying the -a and -o binary primaries and the '(' and ')' operators have been marked obsolescent. 

https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html
Image

bin/runmigration
```diff
@@ -42,27 +40,28 @@ wait_postgres.sh
...
+    # useful when odoo/VERSION was edited
+    if [ -n "$MIG_LOAD_VERSION_CEIL" ]; then
       # Filter dumps of higher releases
-      export MAX_DUMP="$CACHE_DIR/odoo_sample_${MIG_LOAD_VERSION_CEIL}.dmp"
-      CACHED_DUMP=$(ls -v $CACHE_DIR/odoo_sample_*.dmp || true | awk '$0 < ENVIRON["MAX_DUMP"]' | tail -n1)
-    fi
-    if [ -n "$CACHED_DUMP" ]; then
-      echo "No cached migration sample dump found"
+      MAX_DUMP="$CACHE_DIR/odoo_sample_${MIG_LOAD_VERSION_CEIL}.dmp"
+      CACHED_DUMP=$(ls -v $CACHE_DIR/odoo_sample_*.dmp || true | MAX_DUMP=$MAX_DUMP awk '$0 <= ENVIRON["MAX_DUMP"]' | tail -n1)
```
divad1196: `$( { ls -v $CACHE_DIR/odoo_sample_*.dmp || true } | ... )`

to ensure that `ls` always returns true, the `|| true` part needs to be encapsulated in a subshell

bin/runmigration
```diff
@@ -42,27 +40,28 @@ wait_postgres.sh
...
+    else
+      CACHED_DUMP=$(ls -v $CACHE_DIR/odoo_sample_*.dmp || true | tail -n1)
```
divad1196: Same isssue here with `ls || true`

bin/runmigration
divad1196: awk '$0 <= ENVIRON["MAX_DUMP"]' compares strings, not numbers

EDIT: I was partially wrong. `awk` would compare actual numbers, but here we compare versions like `x.y.z` so it does not work.
We can use `dpkg --compare-versions "$(...)" le "$MAX_DUMP"` for version comparison

divad1196 (commented):
Done a quick review of isolate parts and caught a few bugs.

I will review in more depth later

bin/runmigration
florentx: I believe that `|| true` is not needed at all, I've tested a similar script locally

divad1196:
@florentx I finished reviewing all 3 PRs. Except the remarks I already made, the rest looks fine

bin/runmigration
divad1196: @florentx if no file in format `odoo_sample_*.dmp` exist then it will fail. It also happens when `$CACHE_DIR` is undefined/empty because it becomes `/odoo_sample_*.dmp`

We have this case when we have a new cache key after a version bump.
`|| true` is to prevent the failure of `ls` to stop the program due to `set -e`

florentx: comparing strings is not 100% accurate, I see your point:
- "19.0.1.0.11" < "19.0.1.0.5"

Can be improved in the future.

divad1196: @florentx  We can use `dpkg --compare-versions "$(...)" le "$MAX_DUMP"` for version comparison (from my message's edit)

florentx: I've tested, and the exit status of `ls` is swallowed by the pipe, so it does not need the `|| true`.

This is what is in man page:
> the exit status is the exit status of the last command specified in the pipeline

Sample script to test locally:
```sh
set -eux
 
A=$(ls zzzz*|cat)
echo "result A=$A"

B=$(ls zzzz*)
echo "result: B=$B"
```

With output
```sh
~$ /bin/sh test.sh
++ ls 'zzzz*'
++ cat
ls: cannot access 'zzzz*': No such file or directory
+ A=
+ echo 'result A='
result A=
++ ls 'zzzz*'
ls: cannot access 'zzzz*': No such file or directory
+ B=

~$
~$ echo $?
2

```

florentx: except that `ls` returns a list of strings.  It will make the logic and implementation a bit more convoluted.

I prefer to delay it to a separate PR.

florentx: I got something using your trick with `dpkg --compare-versions`.
This should do the job.

This is 2nd commit of the PR: 5fb052c104c258521e2ba944583e3fd73614a000

ivantodorovich (approved):
Thanks!

TBH, I think all these helpers don't belong here. It'd make more sense to have them as separate steps directly in the odoo-ci-workflow github action IMO -- keeping in the image only what's necessary to run the app, and test scripts and helpers in the github actions.

bin/runmigration
divad1196: `set -o pipefail` 
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell

for the test you made: `$()` runs a subshell but you only get the output from it, not the exit code.

also just realized that awk will compare a file `$CACHE_DIR/odoo_sample_*.dmp` to `MAX_DUMP`. I think we miss a grep after ls  to extract the version number.

divad1196:
@ivantodorovich 
> it'd make more sense to have them as separate steps directly in the odoo-ci-workflow

:100: but not with `uses` like we are doing now, this does not work with reusable workflows like we have now.
It needs a ["composite" action](github.action_path).

Instead of chirurgical changes, it's worth planing a redesign. 
=> We should use @florentx changes for now to cleanup the workaround and fix actual bugs

bin/runmigration
divad1196: @florentx cf my last message on the pipeline.
Your test does not represent what the pipeline is doing. The change you made will likely fail on version bump.

florentx: `runmigration` was not run with option `-o pipefail`.  So it explains why I didn't get the failure.

Nevertheless, I've re-added ` || true` fallback.

divad1196: @florentx pipefail is added by github https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell

and your re-added the bug, it runs like `ls || (true | ....)` instead of `(ls || true) | ...`.

PR #478, opened 2026-08-27
fix: remove error message when running runmigration

florentx:
Fix error seen on CI "migration" job:
```
basename: missing operand
Try 'basename --help' for more information.
```

### DIFF
--- a/common/env_odoo.sh
+++ b/common/env_odoo.sh
@@ -8,8 +8,8 @@ USER_ID=${LOCAL_USER_ID:-999}
 id -u odoo &> /dev/null || useradd --shell /bin/bash -u $USER_ID -o -c "" -m odoo
 
 # If run as "gosu odoo ...", shift arguments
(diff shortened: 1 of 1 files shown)


common/env_odoo.sh
```diff
@@ -8,7 +8,7 @@ USER_ID=${LOCAL_USER_ID:-999}
 id -u odoo &> /dev/null || useradd --shell /bin/bash -u $USER_ID -o -c "" -m odoo
 
 # If run as "gosu odoo ...", shift arguments
-if [ "$(basename $1)" = "gosu" ]; then
+if [ "$(basename ${1--})" = "gosu" ]; then
```
divad1196: Fix is ok
`--` looks like a decrement operator, this is a bit confusing to use `-` as the fallback.
In particular, many convention uses `-` as `stdin`

divad1196 (approved):

PR #483, opened 2026-09-10
feat: install libmagic1 for Odoo 19

florentx:
This is required for `python-magic` which is in the base requirements of Odoo 19.


install/package_odoo.sh
```diff
@@ -22,6 +22,11 @@ ODOO_PACKAGE=" \
     xz-utils \
 "
 
+if [ "${VERSION%%.*}" -gt 18 ]; then
+  # Odoo 19+ has dependency on python-magic
+  ODOO_PACKAGE="$ODOO_PACKAGE libmagic1"
+fi
```
ivantodorovich, line 28: Not a fan of having version-specific dependencies on this shared script

Maybe we could have a separated apt-packages.txt file in each version?

gurneyalex (approved):

install/package_odoo.sh
florentx: IMO, it depends how much it diverges. For this single dependency, we're fine.

- there's already a condition below for the `core` image.
- another option could be to install it independently of Odoo version, since it is what we do for most projects, in the odoo-template
- and easier to spot the difference when it's a (short) single file

PR #489, opened 2026-09-15
fix: create missing trigram indexes, with unaccent

florentx:
Fix issue when pg_trgm was not installed before Odoo initialization.

~And add indexes on `"res.partner"` searchable fields.~
- -> PR camptocamp/odoo-tools#31

For this second part, a fix in Odoo is proposed:
- odoo/odoo#288484

---
This is followup of:
- #461


simahawk (commented):
not an expert on the subject but it looks good. 
Are you testing on any existing project locally?

start-entrypoint.d/100_fix_indexes
```diff
@@ -7,20 +7,28 @@ if [ "${ODOO_VERSION%%.*}" -lt 16 -o "${ODOO_VERSION%%.*}" -gt 19 ]; then
...
+expected = [
+    (model._table, field)
+    for model in env
+    if model._auto and not model._abstract
+    for field in model._fields.values()
+    if field.column_type and field.store and field.index == "trigram"
+]
+existing = set()
+
 query = """SELECT indexname, tablename, indexdef FROM pg_indexes $WHERE_CLAUSE;"""
```
gurneyalex: idea: speed up things by adding "AND indexname LIKE '%_index'" -> will remove the pkeys and unique.

gurneyalex (approved):

florentx:
> Are you testing on any existing project locally?

I've tested it with a local copy of a project, Odoo 16, and fixed some issues with the script.

~Also I've extended it to add indexes on res.partner (searched) fields.~
Moved to:
- camptocamp/odoo-tools#31

simahawk (approved):
Hard to review to be honest but LG.
Maybe some more comments would help.
In particular for the part related to res.partner.

PR #490, opened 2026-09-16
fix: runtests, install missing dependencies on dump cache hit

ivantodorovich:
## Problem

CI keys the test DB dump on `git submodule status | md5sum` only. The dump holds the dependencies of the local modules as they were when it was built.

On a cache hit, `runtests` goes straight to the `-i ${LOCAL_ADDONS} --test-enable` pass. When a PR adds a new dependency to a local module, that dependency is not in the dump, so Odoo installs it inside the tests-enabled pass and runs its tests.

This surfaced after 8d3dc55 (do not reinstall add-ons when dump is loaded): before, the untested dependencies pass always ran and silently installed the new dependencies.

## Changes

- On a cache hit, compute the dependencies missing from the restored DB (`ir_module_module` vs `list_dependencies.py`) and install them without tests before the local pass. Costs nothing when nothing changed.


ivantodorovich:
Hello @florentx , do you know why CI is failing on **Scan Image**?
```
Run sysdiglabs/scan-action@v6
Error: Sysdig Secure Token is required for standard execution, please set your token or remove the standalone input.
Error: Sysdig Secure Token is required for standard execution, please set your token or remove the standalone input.
```

ivantodorovich:
Tested here: https://github.com/camptocamp/allchemet_odoo/actions/runs/35115251014/job/104860380831?pr=478 (with https://github.com/camptocamp/allchemet_odoo/pull/478/changes/ee1b7045efef2e2da2a952e4d2894424ff120c41)

florentx:
> Hello @florentx , do you know why CI is failing on **Scan Image**?
> 
> ```
> Run sysdiglabs/scan-action@v6
> Error: Sysdig Secure Token is required for standard execution, please set your token or remove the standalone input.
> Error: Sysdig Secure Token is required for standard execution, please set your token or remove the standalone input.
> ```

It happens because the branch is not in `camptocamp` repo, it is a fork. And token is probably not available in such case.

florentx:
I wonder if it would better fit in `odoo-ci-workflows`

Downsides with the change in `runtests`:
- the install of the additional modules on top of existing ones will happen on each execution, as long as submodules are not upgraded
- it does not support removal of a dependency
- it needs every project to upgrade `docker-odoo-project` to benefit the fix
- and it adds a bit more of CI specific code into `runtests`

So a different approach could be:
- to compute `SUBS_MD5` on both submodules + installed modules:
  something like
  ```sh
  echo "SUBS_MD5=$({ git submodule status; list_dependencies.py; } | md5sum | cut -d ' ' -f1)" >> $GITHUB_OUTPUT
  ```
- so cache will be discarded and re-created if a dependency is added or removed.
- and it is immediately available for all projects, no need to update `docker-odoo-project`

ivantodorovich:
Indeed that sounds much better. I'll rework it tomorrow, closing this for now

PR #444, opened 2026-04-29
fix: multiple CVEs in Werkzeug

florentx:
Multiple CVE reported:
- CVE-2023-25577
- CVE-2024-34069
- CVE-2024-49767
- CVE-2023-46136

Odoo 17 and more recent are not impacted (already have the fixed version)

PR #491, opened 2026-09-17
fix: correct Werkzeug pin for Odoo 16

vvrossem:
## Problem

`16.0/base_requirements.txt` pins `Werkzeug==3.0.6` for `python_version > 3.9`. That's correct for `17.0`/`18.0`/`19.0` (which do support Werkzeug 3.x), but Odoo 16's own `requirements.txt` needs:

```
Werkzeug==0.16.1 ; python_version <= '3.9'
Werkzeug==2.0.2 ; python_version > '3.9'  # (Jammy)
```

`14.0/base_requirements.txt` already correctly pins `Werkzeug==2.0.2` (unconditionally). `16.0`'s file appears to have been copy-pasted from a newer version's pin without adjusting for Odoo 16's actual requirement.

## Impact

Werkzeug 3.x removes APIs Odoo 16 core and some addons still use — e.g. `werkzeug.urls.url_unquote`. Found this while debugging a base image bump (`16.0-4.5.6` → `16.0-5.4.1`) on a downstream project that got stuck failing `test`/`migration` CI with module-loading errors, and separately blocked for a month by a paid module's `ImportError: cannot import name 'url_unquote' from 'werkzeug.urls'`.

## Fix

Mirror Odoo 16's own two-line conditional pin exactly, same pattern already used for other version-specific pins in this file (Jinja2, lxml, psycopg2, etc.).

## Verification

Against `ghcr.io/camptocamp/docker-odoo-project:16.0-5.5.1`:
- `pip install -r base_requirements.txt` with the fix applied installs cleanly, no conflicts.
- Werkzeug lands at `2.0.2`.
- `from werkzeug.urls import url_unquote` succeeds (previously `ImportError` on 3.0.6).


florentx (changes requested):
Actually this risk was detected early.

I've added a note in the `HISTORY` and warned when we released 5.4.1
https://github.com/camptocamp/docker-odoo-project/blob/main/HISTORY.rst#541-2026-08-03

TL;DR;  you need to upgrade Odoo source code + Enterprise, or to pickup a few changes from upstream (see notes).

florentx:
This was:
- PR #444 
- and this comment: https://github.com/camptocamp/docker-odoo-project/pull/444#issuecomment-4656948868

vvrossem:
@florentx would bumping to another version be an alternative ? I've just merged https://github.com/camptocamp/digigram_odoo/pull/241 :/

PR #494, opened 2026-09-21
ci: push to GHCR with the ephemeral GITHUB_TOKEN instead of a PAT

ivantodorovich:
Use the ephemeral token instead of a PAT

Follow-up after merge:

- delete the now unused `GHCR_USER` and `GHCR_TOKEN` secrets


ivantodorovich:
I forgot that we can't use forks here. I'll create a new PR

